Issue #159, opened 2026-02-19
[NIM] Support running NIM models on top of Kserve + Knative


oleksii-donets:
### Name and Version

0.14.0

### What is the problem this feature will solve?

Currently, we are deploying and running NVIDIA Inference Microservices (NIM) in standalone mode. While functional, this standalone setup lacks advanced, request-based autoscaling capabilities—most notably the ability to scale to zero when there is no incoming traffic. Because the models cannot scale down during idle periods, we are constantly reserving compute resources (such as expensive GPUs), leading to inefficient resource utilization and unnecessarily high infrastructure costs.

### What is the feature you are proposing to solve the problem?

Integrate NIM with our existing KServe and Knative serving stack. By wrapping or deploying NIM containers within KServe/Knative, we can leverage our standard serverless infrastructure to manage NIM workloads.

Specifically, this integration will provide:
- Scale-to-Zero capability: Automatically spinning down the NIM pods when traffic stops, freeing up compute resources.
- Traffic-based autoscaling: Dynamically scaling the number of NIM replicas up and down based on real-time request volume.
- Unified Infrastructure: Bringing NIM deployments into our existing operational paradigm, simplifying maintenance, monitoring, and networking.

PR #292, opened 2026-04-20
feat(nim):  use knative scaling for NIM deployments, add configurable storage size for NIM deployments

oleksii-donets:
### Applicable issues

#159 

### Description of changes

- Migrate NIM CRD generation from standalone to kserve inference platform with Knative autoscaling
- Add configurable `storageSize` field to NIM deployment API (create/update/get), replacing the hardcoded 20Gi PVC default
- Add `@ValidStorageSize` constraint backed by Fabric8 `Quantity` parser — accepts Kubernetes quantity formats (binary/decimal suffixes, plain bytes) with configurable upper bound (default 200Gi)
- Add `KubernetesQuantityParser` utility, `StorageSizeValidator`, DB migration V1.57, and comprehensive tests
- Emit explicit `autoscaling.knative.dev/class=kpa.autoscaling.knative.dev` and `autoscaling.knative.dev/metric=concurrency` annotations on all Knative Services when an `ACTIVE_REQUESTS` scaling strategy is provided (previously inherited from cluster defaults). This makes scaling behavior deterministic across clusters and avoids silent drift when operators change cluster-level defaults.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-admin-deployment-manager-backend/actions/runs/24670997942)

ai-dial-actions:
<h1>Dependency Review</h1>
None

.specify/feature.json
```diff
@@ -0,0 +1,3 @@
+{
```
Pasichniuk, line 1: Shouldn't this file be local-only?

specs/015-nim-kserve-migration/checklists/requirements.md
```diff
@@ -0,0 +1,37 @@
+# Specification Quality Checklist: NIM KServe Migration & Configurable Storage Size
```
Pasichniuk: constitution.md states that no per-feature checklists must be generated. Why it happened? Haven't seen them in my branches since constitution was updated

src/main/java/com/epam/aidial/deployment/manager/web/validation/StorageSizeValidator.java
```diff
@@ -0,0 +1,55 @@
...
+            context.buildConstraintViolationWithTemplate("Storage size must be a positive value")
+                    .addConstraintViolation();
+            return false;
+        }
+
+        if (maxStorageSize != null) {
+            BigDecimal maxBytes = KubernetesQuantityParser.parseToBytes(maxStorageSize);
+            if (maxBytes == null) {
+                log.warn("Invalid max storage size configuration: '{}'. Skipping upper bound check.", maxStorageSize);
+                return true;
```
Pasichniuk: Why don't handle invalid configuration as failed validation?

src/main/java/com/epam/aidial/deployment/manager/service/manifest/DeployableManifestGenerator.java
```diff
@@ -22,6 +26,36 @@ public DeployableManifestGenerator(AppProperties appconfig) {
...
+        }
+
+        if (scaling.getStrategy() == null) {
+            return;
+        }
+
+        if (scaling.getStrategy().getType() == ScalingStrategyType.ACTIVE_REQUESTS) {
+            annotations.put(KnativeAnnotations.AUTOSCALING_CLASS, "kpa.autoscaling.knative.dev");
+            annotations.put(KnativeAnnotations.AUTOSCALING_METRIC, "concurrency");
+            annotations.put(KnativeAnnotations.AUTOSCALING_TARGET, String.valueOf(scaling.getStrategy().getThreshold()));
```
Pasichniuk: The new shared helper applyScalingAnnotations unconditionally writes three annotations when the strategy is ACTIVE_REQUESTS:                                             
  annotations.put(AUTOSCALING_CLASS,  "kpa.autoscaling.knative.dev");                                                                                                      
  annotations.put(AUTOSCALING_METRIC, "concurrency");                                                                                                                      
  annotations.put(AUTOSCALING_TARGET, String.valueOf(...));                                                                                                                
                                                                                                                                                                           
  Before this refactor, KnativeManifestGenerator.applyScaling only emitted AUTOSCALING_TARGET (plus containerConcurrency on the revision spec). The class/metric annotations were never written for plain Knative Services.                                                                                                               
                  
  After the refactor, every Knative Service produced by the app will now carry explicit autoscaling.knative.dev/class and autoscaling.knative.dev/metric annotations. Knative usually infers these from cluster defaults, so in most clusters this is harmless — but in clusters where the operator has picked a different default (e.g. HPA class), the new explicit annotation overrides it.                                                                                                                        
                  
  This change isn't mentioned in the PR description, the 015 spec only discusses NIMService, and there is no KnativeManifestGeneratorTest assertion covering it.           
   
  Action: either scope the class/metric emission to NimManifestGenerator (keep Knative byte-identical to development), or add an FR + Knative test documenting the new  behavior intentionally.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/KnativeManifestGenerator.java
Pasichniuk: applyScalingAnnotations already throws with the same message for any non-ACTIVE_REQUESTS strategy, so the else branch is unreachable. Having two places throw identical errors for the same condition is a refactor hazard — if someone later adds a new supported strategy in the shared helper, they'll forget to update this branch.
                                                                                                                                                                           
  Action: simplify to a single if + body; drop the else throw.

src/test/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGeneratorTest.java
```diff
@@ -421,16 +326,180 @@ void testServiceConfig_withoutProbe_setsFallbackProgressDeadlineAnnotation() {
...
+        var generatedService = manifestGenerator.serviceConfig(
+                deploymentName, DM_PREFIX + deploymentName, Collections.emptyList(), Collections.emptyList(), resources, imageName,
+                8000, null, null, scaling, null, STARTUP_TIMEOUT_SEC, null, null
+        );
+
+        // Then: min/max/initial-scale from Scaling + metric/target from strategy
+        var annotations = generatedService.getMetadata().getAnnotations();
+        assertThat(annotations)
+                .containsEntry(KnativeAnnotations.AUTOSCALING_METRIC, "concurrency")
+                .containsEntry(KnativeAnnotations.AUTOSCALING_TARGET, "10")
```
Pasichniuk: shouldSetAutoscalingAnnotations_whenScalingWithStrategyProvided is the only test hitting the ACTIVE_REQUESTS branch, but it only asserts AUTOSCALING_METRIC and AUTOSCALING_TARGET. A regression that drops AUTOSCALING_CLASS would pass tests.                                                                                             
                                                                                                                                                                           
  Action: add .containsEntry(KnativeAnnotations.AUTOSCALING_CLASS, "kpa.autoscaling.knative.dev").

src/main/java/com/epam/aidial/deployment/manager/service/manifest/DeployableManifestGenerator.java
```diff
@@ -22,6 +26,36 @@ public DeployableManifestGenerator(AppProperties appconfig) {
         super(appconfig);
     }
 
+    protected void applyScalingAnnotations(String name, Scaling scaling, Map<String, String> annotations) {
+        var initialScale = Math.max(scaling.getMinReplicas(), 1);
+        annotations.put(KnativeAnnotations.INITIAL_SCALE, String.valueOf(initialScale));
+        annotations.put(KnativeAnnotations.MIN_SCALE, String.valueOf(scaling.getMinReplicas()));
+        annotations.put(KnativeAnnotations.MAX_SCALE, String.valueOf(scaling.getMaxReplicas()));
```
Pasichniuk, line 33: Scaling.maxReplicas is a primitive int. When scaling is non-null but maxReplicas was never set, it defaults to 0. The helper then unconditionally writes  autoscaling.knative.dev/max-scale: "0", overwriting the template default (1). max-scale: 0 is not a valid Knative scaling config.                                        
                                                                                                                                                                           
  The 015 spec edge-case line ("System uses the configured default value for max-scale when maxScale is not set") is only true when scaling == null — not when scaling exists but maxReplicas is unset.
                                                                                                                                                                           
  Action: either skip the MAX_SCALE write when maxReplicas <= 0 (treat 0 as unset), or enforce maxReplicas >= 1 at the validation layer (ScalingDto), and fix the spec wording either way.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/DeployableManifestGenerator.java
```diff
@@ -22,6 +26,36 @@ public DeployableManifestGenerator(AppProperties appconfig) {
...
+            log.trace("Set scale-to-zero-pod-retention-period={} for deployment '{}'", delayStr, name);
+        }
+
+        if (scaling.getStrategy() == null) {
+            return;
+        }
+
+        if (scaling.getStrategy().getType() == ScalingStrategyType.ACTIVE_REQUESTS) {
+            annotations.put(KnativeAnnotations.AUTOSCALING_CLASS, "kpa.autoscaling.knative.dev");
+            annotations.put(KnativeAnnotations.AUTOSCALING_METRIC, "concurrency");
```
Pasichniuk: "kpa.autoscaling.knative.dev" and "concurrency" are Knative-defined values that now appear in code and will likely reappear in tests. Consider KnativeAnnotations.AUTOSCALING_CLASS_KPA / AUTOSCALING_METRIC_CONCURRENCY constants, mirroring the existing annotation-key constants in that class.

.specify/feature.json
oleksii-donets: I haven't found any documentation stating that this file shouldn't be committed. It seems useful to include, especially if another developer needs to take over a PR.

specs/015-nim-kserve-migration/checklists/requirements.md
oleksii-donets: The skills and the constitution contradict each other. Seemingly, the speckit skills should be modified instead of the constitution. However, this is not a good option either, because it would make future updates to speckit error-prone.

oleksii-donets: Agreed — the constitution is authoritative. Removed the per-feature checklist in 68d2ec2.

src/main/java/com/epam/aidial/deployment/manager/web/validation/StorageSizeValidator.java
oleksii-donets: Good catch. Changed in 68d2ec2 — the validator now parses `app.validation.resources.max-storage-size` once in `ConstraintValidator.initialize()` and throws `IllegalStateException` if unparseable, surfacing the misconfig at startup instead of silently bypassing the upper-bound check. Parsed value is cached, so `isValid` no longer re-parses on every call. Added a dedicated test covering the init-time failure.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/DeployableManifestGenerator.java
oleksii-donets: This is intentional — the goal of the shared helper is to move *away* from relying on cluster defaults, which are prone to silent drift (e.g. a cluster switched to HPA class would change our scaling behavior without any code change). Emitting `autoscaling.knative.dev/class=kpa` and `autoscaling.knative.dev/metric=concurrency` explicitly makes the scaling semantics deterministic across clusters, consistent with FR-005 in the 015 spec.

Updated the PR description to call this out, and added a new test `testServiceConfig_withActiveRequestsStrategy_setsExplicitAutoscalingAnnotationsAndConcurrency` in `KnativeManifestGeneratorTest` that asserts all three annotations plus `containerConcurrency` on the revision spec. That codifies the new behavior and prevents accidental regression.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/KnativeManifestGenerator.java
oleksii-donets: Dropped the dead `else` branch in 68d2ec2 — `applyScaling` now just has a single `if` for `ACTIVE_REQUESTS`; the shared helper handles unsupported strategies.

src/test/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGeneratorTest.java
oleksii-donets: Done in 68d2ec2 — the assertion chain now also includes `.containsEntry(AUTOSCALING_CLASS, AUTOSCALING_CLASS_KPA)` (using the new constant introduced for the string-literal comment).

src/main/java/com/epam/aidial/deployment/manager/service/manifest/DeployableManifestGenerator.java
oleksii-donets: `ScalingDto.maxReplicas` carries `@Min(1)` (see `src/main/java/.../web/dto/ScalingDto.java:18-19`), so any request with `maxReplicas < 1` is rejected at the API boundary before it ever reaches the manifest generator. The internal `Scaling` model is only constructed from a validated `ScalingDto` (via `ScalingDtoMapper`) or from persistence — which itself was written through the same validated DTO. In practice there's no path to reach `applyScalingAnnotations` with `maxReplicas == 0` on a non-null `scaling`, so I'd prefer not to add a defensive skip that would hide a programming error if it ever did happen.

The spec wording you quoted is about the case when the `scaling` object itself is null (no scaling configuration on the deployment) — in that case no `min/max/initial-scale` annotations are written and the Knative template defaults apply. Happy to reword the spec if it's still misleading — let me know.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/DeployableManifestGenerator.java
oleksii-donets: Extracted `AUTOSCALING_CLASS_KPA` and `AUTOSCALING_METRIC_CONCURRENCY` in `KnativeAnnotations` in 68d2ec2. Both tests (`NimManifestGeneratorTest` and the new `KnativeManifestGeneratorTest` case) reference the constants as well.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

PR #296, opened 2026-04-22
feat: stop image build

### LINKED ISSUE #138: [Deployments -> Images] Add 'Stop' button to interrupt installation process
As a user, I want to be able to stop an image installation in progress,
so that I can interrupt it if needed without waiting for it to complete.

### Expected behavior
- A "Stop" button is visible during image installation
- Clicking "Stop" interrupts the installation process

### Current behavior
There is no way to interrupt the installation once it has started.

### Confidential information

- [x] I confirm that do not share any confidential information

Pasichniuk:
### Applicable issues

- fixes #138 

### Description of changes

  - Feature 017 — administrators can stop an in-progress image build. A new DELETE /{id} endpoint lets a full admin interrupt a running build; the cluster-side Job and its pods are torn down, cluster capacity is released, and the image definition lands in a new terminal status BUILD_STOPPED (distinct from success/failure) so the admin can immediately retrigger a corrected build. Logs captured up to the interruption are preserved, and any open status/log streams close cleanly with a final event.          
  - Safe behavior around races and retries. Stop only applies to builds actually in progress; requests for terminal, never-built, or unknown definitions are rejected with a clear error that echoes the build's current status so UIs can reconcile without a refresh. Concurrent stop-vs-natural-completion is handled deterministically (last committer wins, under a row lock), and repeat stops are idempotent. If cluster-side cleanup fails, the whole operation fails and the status stays BUILDING so nothing drifts between DB and cluster.                                                                                                                                           
  - Configurable stop timeout. A new setting controls how long to wait for Kubernetes to confirm Job/pod deletion before giving up, tunable per environment.
  - Follow-up hardening in the build pipeline. Build Jobs are now labeled with the image definition so they can be discovered and deleted reliably, and the pipeline tolerates a Job that's been removed out from under it (externally deleted) without crashing the worker — it fails the build gracefully instead. Timeout/namespace settings were consolidated into a single place in the pipeline runner, removing duplication across the four build steps.                                                 
  - Tests and spec artifacts. Added an end-to-end StopImageBuildFunctionalTest that runs across all three DB vendors (H2, Postgres, SQL Server), plus service-, controller-, and k8s-client-level coverage for the stop flow and the externally-deleted-Job case. The full speckit bundle (spec, plan, tasks, research, data model, API contract, quickstart) is checked in under specs/017-stop-image-build/.

### Checklist

- [ ] Title of the pull request follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/)


ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-admin-deployment-manager-backend/actions/runs/24768810964)

src/main/java/com/epam/aidial/deployment/manager/web/controller/ImageBuildController.java
```diff
@@ -45,6 +48,13 @@ public void buildImage(@RequestBody @Valid CreateBuildImageRequestDto requestDto
         imageBuildRunner.buildImage(requestDto.imageDefinitionId());
     }
 
+    @FullAdminOnly
+    @DeleteMapping(path = "/{id}")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
```
oleksii-donets, line 59: **Missing OpenAPI annotations**

The other endpoints in this controller have `@Operation(summary = ...)` and `@ApiResponse` annotations (constitution §API Conventions). The task T019 also explicitly calls for them. Please add:

```java
@Operation(summary = "Stop an in-progress image build")
@ApiResponse(responseCode = "204", description = "Build stopped successfully")
@ApiResponse(responseCode = "400", description = "Build is not in progress or cluster-side deletion failed")
@ApiResponse(responseCode = "404", description = "Image definition not found")
```

src/main/java/com/epam/aidial/deployment/manager/dao/repository/ImageDefinitionRepository.java
```diff
@@ -126,7 +126,7 @@ public void startBuild(UUID id) {
     }
 
     public void completeBuildSuccessfully(UUID id, String imageName, long builtAt) {
```
oleksii-donets, line 128: **Add a code comment explaining why there's no `BUILD_STOPPED` guard here**

The research doc (R1) explicitly and deliberately keeps `completeBuildSuccessfully` unconditional — if the pipeline's Job had already succeeded, the artifact is adopted as `BUILD_SUCCESSFUL` rather than orphaned. This is a narrow-race trade-off.

Without a comment, a future contributor will see `failBuild` has a `BUILD_STOPPED` guard and "fix" this method to match — breaking the intended behavior. Something like:

```java
// Intentionally unconditional: if the pipeline's Job succeeded before the stop committed,
// the built artifact is adopted as BUILD_SUCCESSFUL (see spec Edge Case "Stop requested as the build is completing")
```

src/main/java/com/epam/aidial/deployment/manager/dao/mapper/PersistenceImageDefinitionViewMapper.java
```diff
@@ -74,6 +74,7 @@ private ImageStatus mapImageStatus(PersistenceImageStatus status) {
             case BUILDING -> ImageStatus.BUILDING;
             case BUILD_FAILED -> ImageStatus.BUILD_FAILED;
             case BUILD_SUCCESSFUL -> ImageStatus.BUILD_SUCCESSFUL;
+            case BUILD_STOPPED -> ImageStatus.BUILD_STOPPED;
```
oleksii-donets, line 77: nit: Missing trailing newline at end of file.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/ImageBuildFromGitPipeline.java
```diff
@@ -37,12 +38,19 @@ public void run(UUID imageDefinitionId) {
         var imageDefinition = imageDefinitionService.getImageDefinition(imageDefinitionId)
                 .orElseThrow(() -> new EntityNotFoundException("Image definition not found by id: %s".formatted(imageDefinitionId)));
 
+        boolean externallyInterrupted = false;
```
oleksii-donets, line 41: **Suggestion: extract the duplicated externally-interrupted pattern**

This exact `boolean externallyInterrupted` + try/catch/finally pattern is now repeated verbatim in all three pipelines (`ImageBuildFromGitPipeline`, `ImageCopyPipeline`, `ImageWrapperBuildPipeline`).

Consider extracting to a template method on a shared base, or a utility that accepts a `Runnable` for the pipeline action and the error-message prefix. Not blocking — the duplication is clear and each pipeline is short — but it would prevent drift if the pattern needs updating later.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/K8sClient.java
```diff
@@ -158,6 +160,36 @@ public void deleteJob(String namespace, String name) {
...
+        }
+        for (Job job : existing) {
+            var name = job.getMetadata().getName();
+            client.batch().v1().jobs()
+                    .inNamespace(namespace)
+                    .withName(name)
+                    .waitUntilCondition(Objects::isNull, timeoutSec, TimeUnit.SECONDS);
+        }
+        log.debug("All Jobs matching label {}={} in namespace '{}' fully deleted.", labelKey, labelValue, namespace);
+    }
```
oleksii-donets, line 178: nit: The timeout is applied per-Job (`waitUntilCondition` for each), so with N matching Jobs the worst-case total wait is `N × timeoutSec`. For this use case N=1, so it's fine, but worth a brief javadoc note mentioning "timeout is per-Job, not aggregate" for future callers.

src/main/java/com/epam/aidial/deployment/manager/service/ImageBuildStopService.java
```diff
@@ -0,0 +1,68 @@
...
+        if (currentStatus != ImageStatus.BUILDING) {
+            throw new ImageBuildNotInProgressException(imageDefinitionId, currentStatus);
+        }
+
+        try {
+            jobRunner.deleteJob(imageDefinitionId);
+        } catch (Exception cause) {
+            log.warn("Failed to delete build Job for image definition '{}': {}", imageDefinitionId, cause.getMessage(), cause);
+            throw new ImageBuildStopFailedException(cause);
+        }
```
oleksii-donets, line 44: Good: The orchestration flow is clean and well-ordered — validate status, delete K8s resources, clean disposables, post-condition check, then DB status flip. The all-or-nothing semantics (FR-004) are correctly enforced.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/JobRunner.java
```diff
@@ -66,16 +75,20 @@ public boolean run(
         }
 
         var job = jobSpecification.getJob();
+        tagJobWithGroupId(job, groupId);
         disposableResourceManager.saveK8sResources(List.of(job), K8sResourceKind.JOB, groupId, namespace);
         client.createJob(namespace, job);
         log.info("Created build job. jobId: '{}'", jobId);
         jobCallback.onJobPhaseChange(JobPhase.CREATED);
```
oleksii-donets, line 82: Good: Making the `waitJob` predicates null-safe (`j == null || ...`) and throwing `JobExternallyDeletedException` when the wait returns null is a clean approach. The pipeline threads can now distinguish "my Job failed" from "my Job was deleted out from under me" without any shared state.

oleksii-donets (commented):
See inline comments. Overall a well-structured feature — the race-safety analysis and Envers compatibility are thorough.

src/main/java/com/epam/aidial/deployment/manager/dao/repository/ImageDefinitionRepository.java
```diff
@@ -136,18 +136,40 @@ public void completeBuildSuccessfully(UUID id, String imageName, long builtAt) {
...
 
     public void failBuild(UUID id, String errorLog) {
-        var entity = findImageDefinitionById(id);
+        var entity = findImageDefinitionForUpdate(id);
+        if (entity.getBuildStatus() == PersistenceImageStatus.BUILD_STOPPED) {
+            log.debug("Skipping build failure for image definition '{}': build was stopped by administrator", id);
+            return;
+        }
         entity.setBuildStatus(PersistenceImageStatus.BUILD_FAILED);
         imageDefinitionJpaRepository.saveAndFlush(entity);
```
oleksii-donets: **Redundant `saveAndFlush` on already-managed entities**

The entity is loaded via `findImageDefinitionForUpdate` (`SELECT ... FOR UPDATE`), so it's already managed by the persistence context. After `setBuildStatus(...)`, Hibernate dirty-tracks it and will auto-flush at transaction commit. The explicit `saveAndFlush` is redundant here.

For `failBuild` specifically, there are two flush round-trips within the same transaction:
1. `imageDefinitionJpaRepository.saveAndFlush(entity)` — flushes status
2. `appendBuildLogs(...)` → `imageBuildLogsJpaRepository.saveAndFlush(logsEntity)` — flushes logs

Since both are in the same `@Transactional` boundary, early flushing provides no durability benefit (a rollback undoes all flushes). You could remove the explicit `saveAndFlush` on the status entity and let dirty checking + commit-time flush handle it, reducing unnecessary DB round-trips.

Same applies to `stopBuild` and `completeBuildSuccessfully`.

src/main/java/com/epam/aidial/deployment/manager/web/controller/ImageBuildController.java
Pasichniuk: Added

src/main/java/com/epam/aidial/deployment/manager/dao/repository/ImageDefinitionRepository.java
Pasichniuk: Added

src/main/java/com/epam/aidial/deployment/manager/dao/mapper/PersistenceImageDefinitionViewMapper.java
Pasichniuk: Added

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/ImageBuildFromGitPipeline.java
Pasichniuk: Will skip. Each pipeline is short, duplication is obvious. Matches the "don't introduce abstractions beyond what the task requires" guidance.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/K8sClient.java
Pasichniuk: Added a note

src/main/java/com/epam/aidial/deployment/manager/dao/repository/ImageDefinitionRepository.java
Pasichniuk: Fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/web/controller/ImageBuildController.java
```diff
@@ -45,6 +50,17 @@ public void buildImage(@RequestBody @Valid CreateBuildImageRequestDto requestDto
         imageBuildRunner.buildImage(requestDto.imageDefinitionId());
     }
 
+    @Operation(summary = "Stop an in-progress image build")
+    @ApiResponse(responseCode = "204", description = "Build stopped successfully")
+    @ApiResponse(responseCode = "400", description = "Build is not in progress or cluster-side deletion failed")
+    @ApiResponse(responseCode = "404", description = "Image definition not found")
+    @FullAdminOnly
+    @DeleteMapping(path = "/{id}")
```
oleksii-donets, line 58: **`DELETE` is the wrong HTTP method for a stop/cancel action**

`DELETE /api/v1/images/builds/{id}` semantically means "delete the build resource," but the record survives with `BUILD_STOPPED` — logs preserved, status persisted, audit recorded. Nothing is deleted from the API consumer's perspective.

Consider `POST /api/v1/images/builds/{id}/stop` instead — an explicit action endpoint with no ambiguity about what happens to the resource. This matches the pattern used by cloud providers for cancelling async operations (e.g. GCP `operations/{id}:cancel`).

`DELETE` would be appropriate if the build record itself were being removed (like the existing `DELETE /api/v1/images/{id}` for image definitions). Here the record survives — that's a state transition, not a resource deletion.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/JobRunner.java
```diff
@@ -66,16 +75,20 @@ public boolean run(
...
 
         log.debug("Waiting for job to start. jobId: '{}'", jobId);
-        Predicate<Job> jobIsRunning = j -> JobPhase.fromJob(j)
+        Predicate<Job> jobIsRunning = j -> j == null || JobPhase.fromJob(j)
```
oleksii-donets, line 85: **`j == null` in wait predicates is unsafe — K8s is eventually consistent**

The predicates treat `null` as "condition met" (done waiting). But `waitUntilCondition` relies on Kubernetes' asynchronous event system, not a direct etcd query. After `createJob` returns 201, there is a propagation window:

1. etcd notifies the API Server that a new Job was added
2. The API Server updates its internal Watch Cache
3. The API Server broadcasts an "Added" event over the network to listeners (including the Fabric8 client)
4. The Fabric8 client receives the event and updates its local state

If `waitUntilCondition` evaluates the predicate before step 4 completes, the Job isn't visible yet — the predicate receives `null`, treats it as "done," and the method throws `JobExternallyDeletedException` while the Job is actually running fine on the cluster. The pipeline then skips `failBuild`, leaving the status stuck in `BUILDING` forever.

Please remove the `j == null ||` from both predicates and verify how Fabric8's `waitUntilCondition` actually signals resource deletion vs. not-yet-visible. Find a safer way to detect external deletion that doesn't conflict with eventual consistency.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/K8sClient.java
```diff
@@ -158,6 +160,40 @@ public void deleteJob(String namespace, String name) {
         log.debug("Job '{}' successfully deleted.", name);
     }
 
+    /**
+     * Deletes all Jobs matching the given label and waits for each to be fully removed.
+     * Note: {@code timeoutSec} is applied per matched Job, not as an aggregate — worst-case total wait is N × timeoutSec.
+     */
+    public void deleteJobsByLabelAndWait(String namespace, String labelKey, String labelValue, int timeoutSec) {
```
oleksii-donets, line 166: **Simplify: replace two loops with a single Fabric8 call**

The two-loop approach (N delete calls + N wait calls) can likely be replaced with:

```java
client.batch().v1().jobs()
    .inNamespace(namespace)
    .withLabel(labelKey, labelValue)
    .withPropagationPolicy(DeletionPropagation.FOREGROUND)
    .withGracePeriod(0L)
    .withTimeout(timeoutSec, TimeUnit.SECONDS)
    .delete();
```

One call instead of list + N deletes + N waits. `withTimeout` on the delete chain should make Fabric8 block until the server confirms removal. This also fixes the per-Job timeout issue — the timeout applies to the whole operation, not per-Job.

Please verify `withTimeout` on delete blocks until removal is confirmed in Fabric8 7.5.2.

src/main/java/com/epam/aidial/deployment/manager/web/controller/ImageBuildController.java
Pasichniuk: Not wrong, this was a deliberate API decision ratified in plan.md and that the DELETE verb was accepted with full context of what it affects. I agree that record survives the stop, but K8s resources aren't, they are deleted. To be more precise, we delete all K8s resources related to image build and do a status transition for image definition. Both POST and DELETE are reasonable choices, but I'd like to stick with the current approach to avoid refactoring of specs, tests and re-explaining new contract to FE team.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/JobRunner.java
Pasichniuk: Your mental model of Fabric8's waitUntilCondition is partially right (it does use a Watch), but the null-as-deletion semantics are explicitly documented:      
                            
  - research.md:42-44 records the exact failure mode ("pipeline-runner thread blocked inside K8sClient.waitJob(...) unblocks with a null Job") and the decision to make predicates null-safe.     
  - K8sKnativeClient.java:99-101 already uses .waitUntilCondition(Objects::isNull, ...) as the canonical "wait until gone" idiom — so null == deleted is an established convention in this codebase, not an invention of this PR.                                                                                                                
   
The proposed failure scenario (initial GET returning null on a just-created Job because the apiserver Watch Cache lags) requires:                             
  1. client.createJob(...) returns 201 with a resourceVersion at line 80
  2. The immediately following GET inside waitJob at line 88 — on the same Fabric8 client, same apiserver — misses the resource                                            
                                                                                                                               
  K8s provides read-after-write consistency within a single apiserver session, and Fabric8 uses one HTTP client against one endpoint. The race is nearly impossible in single-apiserver dev/test setups and very rare in HA. If it does fire, the Job leaks on the cluster and the DB row stays at BUILDING — undesirable but recoverable (admin can issue stop).

src/main/java/com/epam/aidial/deployment/manager/kubernetes/K8sClient.java
Pasichniuk: - Fabric8's withTimeout(...) configures the HTTP request timeout, not a "wait until the resource is actually gone" timeout.                                              
  - delete() with FOREGROUND propagation returns as soon as the apiserver accepts the request (sets deletionTimestamp + finalizers); the actual removal happens asynchronously via the GC controller.                                                                                                                                    
  - To block until the Job is gone, you still need waitUntilCondition(Objects::isNull, ...) — which is exactly what the current code does and what research.md:87 specifies as canonical ("Do NOT use custom polling loops anywhere; Fabric8's waitUntilCondition is canonical per constitution §Key Patterns").                                    
                            
  The one directionally-valid idea in the suggestion — withLabel(...).delete() as a single bulk call — is a cosmetic micro-optimization (1 DELETE request vs N), but you still need the list for the per-Job wait phase, so you save approximately nothing.
                                                                                                                                                                           
The current two-loop shape (list → delete each → wait each) is correct, and the docstring note you already added ("timeout is per matched Job") addresses the only real concern.

oleksii-donets: The claim that withTimeout() configures the HTTP request timeout is incorrect. Fabric8's own Javadoc for delete() states:

  ▎ To make sure that a resource is completely deleted and not only marked for deletion, use withTimeout(long, TimeUnit) to perform a blocking delete operation.

  Internally, withTimeout().delete() sends the DELETE request, then calls waitForDelete() which creates a SharedIndexInformer (Watch) that blocks until all deleted UIDs are confirmed gone — the same informOnCondition() mechanism that waitUntilCondition uses. So this isn't
  bypassing the canonical Watch pattern; it's using it.

  The suggested single-call approach:
  client.batch().v1().jobs()
      .inNamespace(namespace)
      .withLabel(labelKey, labelValue)
      .withPropagationPolicy(DeletionPropagation.FOREGROUND)
      .withGracePeriod(0L)
      .withTimeout(timeoutSec, TimeUnit.SECONDS)
      .delete();
  would replace list + N deletes + N waits with one bulk DELETE + one aggregate Watch-based wait. The timeout also becomes aggregate rather than per-Job, which fixes the N × timeout worst case noted in the current docstring.

  Worth adopting as a simplification.

Pasichniuk: Thanks for a detailed explanation. Simplified this method.

ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

PR #300, opened 2026-04-24
feat(nim): make kserve/knative serving mode opt-in via env var

### LINKED ISSUE #159: [NIM] Support running NIM models on top of Kserve + Knative

oleksii-donets:
### Applicable issues

- fixes #159

### Description of changes

Restores the pre-#292 ingress-based NIM behavior as the default and gates the
kserve/knative serverless mode behind the new `app.nim.deploy.kserve-mode-enabled`
property (env `K8S_NIM_DEPLOYMENT_KSERVE_MODE_ENABLED`, default `false`), so
existing deployments keep working without configuration changes.

- Legacy mode (default): `inferencePlatform: standalone`, restored `expose.ingress`
  with TLS and nginx rules from `K8S_NIM_CLUSTER_HOST` when
  `use-cluster-internal-url=false`, no `expose.router`, no Knative autoscaling
  annotations. `cluster-host` validation is required only in this mode when
  external URL is used.
- KServe mode (opt-in): `inferencePlatform: kserve`, `expose.router`, and
  Knative autoscaling annotations derived from the deployment `Scaling`.
- `NimDeployProperties` grows `kserveModeEnabled` and `clusterHost` back;
  `AppProperties` restores `nimServiceExposeIngressConfig`; `application.yml`
  restores `nim-service-expose-ingress-config`/`cluster-host` and drops the
  hardcoded kserve/router/autoscaling defaults from `nim-service-config`.
- Tests and JSON fixtures updated; new full-manifest fixtures assert both
  modes end-to-end and cluster-host validation edge cases.
- `docs/configuration.md` and `specs/nim-deployments/spec.md` document the
  flag; `specs/015-nim-kserve-migration/spec.md` annotated as opt-in.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGenerator.java
```diff
@@ -35,16 +44,26 @@
 public class NimManifestGenerator extends DeployableManifestGenerator {
 
     private static final String NIM_SERVED_MODEL_NAME_ENV = "NIM_SERVED_MODEL_NAME";
+    private static final String TLS_SECRET_NAME_SUFFIX = "-tls-secret";
+    private static final String DEFAULT_INITIAL_SCALE = "1";
+    private static final String DEFAULT_MIN_SCALE = "1";
+    private static final String DEFAULT_MAX_SCALE = "1";
+    private static final String DEFAULT_AUTOSCALING_CLASS = "kpa.autoscaling.knative.dev";
+    private static final String DEFAULT_AUTOSCALING_METRIC = "concurrency";
```
Pasichniuk: Reuse constants from KnativeAnnotations

src/main/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGenerator.java
```diff
@@ -129,14 +166,65 @@ private void applyExposeService(MappingChain<Expose> exposeChain, int httpPort,
...
         log.debug("Applying scaling for NIM deployment '{}': {}", name, scaling);
+        var annotations = config.get(NimMappers.SERVICE_METADATA_FIELD)
+                .get(NimMappers.METADATA_ANNOTATIONS_FIELD).data();
         if (scaling == null) {
+            annotations.put(KnativeAnnotations.INITIAL_SCALE, DEFAULT_INITIAL_SCALE);
+            annotations.put(KnativeAnnotations.MIN_SCALE, DEFAULT_MIN_SCALE);
+            annotations.put(KnativeAnnotations.MAX_SCALE, DEFAULT_MAX_SCALE);
+            annotations.put(KnativeAnnotations.AUTOSCALING_CLASS, DEFAULT_AUTOSCALING_CLASS);
+            annotations.put(KnativeAnnotations.AUTOSCALING_METRIC, DEFAULT_AUTOSCALING_METRIC);
+            annotations.put(KnativeAnnotations.AUTOSCALING_TARGET, DEFAULT_AUTOSCALING_TARGET);
```
Pasichniuk: In the updated `applyScaling`:

- `scaling == null` → emits `initial=1`, `min=1`, `max=1`, `class=kpa`, `metric=concurrency`, `target=10`.
- `scaling != null` with `strategy == null` → falls through to `applyScalingAnnotations`, which sets `min/max/initial/scale-to-zero` **but not** `class`, `metric`, or `target`.

The commit message for `ce93ff0` argues the fixed class/metric/target defaults are needed so behavior "does not depend on Knative cluster-level defaults". By that same argument, the `Scaling(min=N, max=M, strategy=null)` branch should also emit them. Currently a caller that passes `new Scaling(1, 1, null, null)` gets materially different behavior from passing `null`, which is surprising.

Two options:
- apply the class/metric/target defaults in `applyScalingAnnotations` when `strategy == null` (makes the shape consistent across both branches), or
- explicitly document and test the asymmetry if it's intentional (e.g., "explicit `Scaling` opts out of class/metric/target defaults").

src/main/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGenerator.java
```diff
@@ -65,6 +84,12 @@ public NIMService serviceConfig(
             @Nullable List<String> args,
             @Nullable Map<String, String> nodePoolLabels
     ) {
+        boolean kserveMode = nimDeployProperties.isKserveModeEnabled();
+        boolean useExternalUrl = !nimDeployProperties.isUseClusterInternalUrl();
+        if (!kserveMode && useExternalUrl && StringUtils.isBlank(nimDeployProperties.getClusterHost())) {
+            throw new IllegalArgumentException("External NIM URL is enabled but cluster host is not configured");
+        }
```
Pasichniuk: The `@AssertTrue isClusterHostValid` on `NimDeployProperties` already prevents the app from starting in this state, so the runtime branch is unreachable in practice. It's cheap defensive coding and acceptable, but mention it in the test comment or drop it — right now readers have to reason about both layers. If keeping it, consider moving the message into a constant shared with the validator so they don't drift.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGenerator.java
oleksii-donets: Done in 87229ea4. `applyScaling` now delegates to the shared `DeployableManifestGenerator.applyScalingAnnotations`, which already uses `KnativeAnnotations.AUTOSCALING_CLASS_KPA` / `AUTOSCALING_METRIC_CONCURRENCY`. All three duplicate local constants (`DEFAULT_AUTOSCALING_CLASS`, `DEFAULT_AUTOSCALING_METRIC`, `DEFAULT_AUTOSCALING_TARGET`) are gone.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGenerator.java
oleksii-donets: Good catch — fixed in 87229ea4 by collapsing both branches into a single call:

```java
applyScalingAnnotations(name, scaling != null ? scaling : DEFAULT_SCALING, annotations);
```

where `DEFAULT_SCALING = new Scaling(1, 1, null, null)`. Both `scaling == null` and `Scaling(min=N, max=M, strategy=null)` now run through the same code path: `min-scale` / `max-scale` / `initial-scale` are always set (defaulting to `1/1/1`), and `class` / `metric` / `target` are set only when a `Scaling` with an `ACTIVE_REQUESTS` strategy is provided — otherwise we defer to Knative's cluster defaults. Specs (`specs/nim-deployments/spec.md` KServe scenario and the `015-nim-kserve-migration` revision note) updated to match.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/NimManifestGenerator.java
oleksii-donets: Agreed, dropped the runtime branch in 87229ea4 along with the test that asserted it — `@AssertTrue NimDeployProperties.isClusterHostValid` remains the single source of truth and fails Spring context startup before any manifest generation runs. Also updated the "Legacy mode rejects missing cluster host" scenario in `specs/nim-deployments/spec.md` to describe startup-validation-only enforcement.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

Issue #303, opened 2026-04-24
[Images] On installed image's description, author, topics, license update - exception it thrown


Pasichniuk:
### Name and Version

ai-dial-admin-deployment-manager-backend [0.15.0]

### What steps will reproduce the bug?

1. Create image definition
2. Run installation
3. After successfully installed, try to update description/author/topics/license

### What is the expected behavior?

Update of meta fields like description/author/topics/license which do not affect image installation is allowed and build-related fields are not changed (built status, image name and logs are kept)

### What do you see instead?

Exception is thrown, not allowing any image definition updates since it has BUILD_SUCCESSFUL status.

PR #304, opened 2026-04-24
feat:  allow meta fields updates of built image definition

Pasichniuk:
### Applicable issues

- fixes to #303

### Description of changes

- New behavior: built image definitions (BUILD_SUCCESSFUL) can now have their description, author, topics, and license updated without invalidating the built image. Any change to build-affecting fields is still rejected, and images currently building remain read-only.                                                                      
- Scope of the change: the "did build inputs change?" check lives on the ImageDefinition domain model (with order-insensitive allowedDomains comparison), the service routes meta-only updates through a dedicated mapper/repository path that preserves build state, unit and functional tests cover the new behavior, and specs/image-definitions/spec.md documents it. 

### Checklist

### DIFF
--- a/specs/image-definitions/spec.md
+++ b/specs/image-definitions/spec.md
@@ -80,13 +80,37 @@ Status: **Implemented**
 - **THEN** the system responds with 400
 
 ### Requirement: Update image definition
-The system SHALL update an existing image definition by ID.
+The system SHALL update an existing image definition by ID. The set of fields that may be mutated depends on the current `buildStatus`:
+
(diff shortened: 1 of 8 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):
LGTM, approving.

One thing worth flagging for follow-up (not blocking): `hasSameBuildAffectingFields` enumerates the *build-affecting* fields, which means a new field added to `ImageDefinition` (or any subtype) defaults to being treated as a meta field. That's the wrong default — silently allowing a `200 OK` on a built image while ignoring the change is exactly the failure mode we hit with the rolling-update detection on `DeploymentService` (new field added, comparison wasn't updated, deployment didn't roll). Fail-closed is safer here too: list the meta fields explicitly and treat everything else as build-affecting, so a forgotten new field produces a loud 400 instead of a silent no-op. Worth a follow-up issue.

PR #319, opened 2026-05-11
feat!: explicit primitives + create-time defaults

### LINKED ISSUE #285: feat: add node pool selection for deployments
### Name and Version

[BE] 0.16.0-rc

Administrators deploying models and other workloads (MCP, Adapter, Interceptor, Application, Inference, NIM) have no way to control which cluster nodes the workload is scheduled on. This leads to:

- GPU workloads landing on CPU-only nodes (or vice versa), causing scheduling failures or wasted expensive resources
- No visibility into available node pools and their hardware capabilities, forcing administrators to cross-reference external cloud dashboards or run kubectl commands
- Manual, error-prone workflow to ensure workloads match the right infrastructure tier

**1. Node pool selection on deployments**
- Add an optional `nodePool` field to all deployment types
- When specified, the system injects the appropriate node selector constraint into the generated Kubernetes manifest (KNative, KServe, NIM)
- When not specified, default Kubernetes scheduling is preserved (backward compatible)
- Support clearing a previously set node pool (setting it back to none)

**2. Node pool discovery endpoint with resource summaries**
- New endpoint to list available node pools discovered from cluster node labels
- Each pool includes a resource summary: pool name, node count (total and ready), instance type(s), GPU type and count per node (if present), and total allocatable CPU, memory, and GPU across the pool
- The node pool label key is configurable via environment variable to support different cloud providers (GKE, AKS, EKS, custom)

**3. Validation**
- Non-blank validation when a node pool value is provided

**Scope boundary:** This feature covers node selector-based placement only. Advanced scheduling (tolerations, node affinity, pod affinity) may be added in a future iteration.

**Spec:** `specs/018-node-pool-selection/spec.md`

- **Node affinity rules**: More expressive (supports preferred/required, operators like In/NotIn), but significantly more complex for the user to configure. Node selector covers the primary use case with minimal UX friction.
- **Tolerations management**: Would allow scheduling on tainted nodes (e.g., dedicated GPU pools), but is a separate concern from pool targeting. Can be layered on top in a follow-up.
- **Cloud provider API integration**: Could provide richer data (e.g., max autoscaling node count per pool), but ties the application to a specific cloud provider. The current approach is cloud-agnostic, relying only on standard Kubernetes node labels and status.

oleksii-donets:
### Applicable issues

- fixes #285

### Description of changes

Replaces Feature 016's derived-label-key + per-pool capacity-numbers configuration with explicit Kubernetes scheduling primitives per pool (`nodeSelector`, `affinity`, `tolerations`) parsed from `NODE_POOLS` as YAML in strict mode. Adds two startup-validated fallback environment variables (`NODE_POOL_DEFAULT`, `NODE_POOL_DEFAULT_MODEL`) that stamp onto new deployments via a create-time cascade.

Key changes:

- **Schema breaking** — `NODE_POOLS` is now a YAML document (JSON-as-YAML subset accepted). Strict-mode parsing rejects Feature 016 legacy fields (`maxNodes`, `cpu`, `memory`, `gpu`). Operators upgrading must rewrite their config; see `docs/configuration.md`.
- **Pool identity split** — each pool carries an immutable `id` (machine identifier; deployments persist this as `node_pool_id`) and a mutable `name` (display label, safe to rename without breaking deployments). `NODE_POOL_DEFAULT` / `NODE_POOL_DEFAULT_MODEL` reference ids. The `defaults` block of the listing endpoint exposes `defaultId` / `modelId`. Deployment responses carry both `nodePoolId` (persisted) and `nodePoolName` (resolved at read time).
- **Manifest projection** — pool primitives are applied verbatim to Knative / NIM / KServe-Inference pod templates. `PoolPrimitivesConverter` bridges Fabric8 model types to the CRD-generated equivalents that NIM and KServe use.
- **Create-time cascade** — when the create payload's `nodePoolId` is null (whether omitted or explicitly null), the system resolves model-override → catch-all → null and **stamps** the resolved value onto the record. The create response returns the stamped value; users see exactly which pool will be used. The decision to apply the cascade lives at the service entry point: public `createDeployment(...)` delegates to a private overload with `applyNodePoolDefaultIfEmpty=true`; duplicate calls the overload directly with `false` so the source's pool is preserved verbatim. No model-level presence flag.
- **Update is PUT-style for `nodePoolId`** — payload value is authoritative; null clears the stored pool; absent is equivalent to null. Callers PATCHing other fields must include the current pool. The cascade never runs on update; admin env-var changes never migrate existing deployments.
- **Listing API** reshaped to `{ pools, defaults }`; fields not declared on a pool are omitted from the response.
- **Export/Import** — `nodePool` is environment-specific. Export strips it; imports run through the target environment's cascade.
- **Tests** — new unit tests for every new component (`WorkloadClassifier`, `PoolPrimitivesConverter`, `DeploymentExportMixIn`); existing configuration / service / controller / manifest tests rewritten for the new schema. `./gradlew testFast` green; checkstyle clean.
- **Docs** — `docs/configuration.md` Node Pool Configuration section trimmed to admin essentials (three-env-var table, one example, migration callout).
- **Spec conventions** — feature spec declares `**Capability**: deployments`; listed in `specs/README.md` in-flight table.

Spec, plan, research, data-model, contracts, quickstart, and tasks are under `specs/018-explicit-pool-scheduling/`.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/configuration/NodePoolConfiguration.java
```diff
@@ -20,42 +22,57 @@
...
     public NodePoolProperties nodePoolProperties(
-            @Value("${app.node-pools.label-key:}") String nodePoolLabelKey,
-            @Value("${app.node-pools.pools:}") String nodePoolsJson,
+            @Value("${app.node-pools.pools:}") String nodePoolsYaml,
+            @Value("${app.node-pools.default:}") String defaultPool,
+            @Value("${app.node-pools.default-model:}") String defaultModelPool,
+            @Value("${NODE_POOL_LABEL_KEY:}") String legacyLabelKey,
             Validator validator) {
 
+        if (StringUtils.isNotBlank(legacyLabelKey)) {
```
oleksii-donets: just remove

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/web/controller/NodePoolController.java
```diff
@@ -26,9 +24,11 @@ public class NodePoolController {
...
     @Operation(summary = "List available node pools",
-            description = "Returns all configured node pools with their hardware specifications")
+            description = "Returns the configured pool catalogue (id, name, optional description). "
+                    + "Scheduling primitives are internal configuration and not exposed via this endpoint. "
+                    + "The `pools` array is always present and is empty when no pools are configured.")
     @ApiResponse(responseCode = "200", description = "Node pools retrieved successfully")
-    public List<NodePoolDto> getNodePools() {
-        return nodePoolDtoMapper.toNodePoolDtoList(nodePoolService.getNodePools());
+    public NodePoolListResponseDto getNodePools() {
+        return nodePoolDtoMapper.toListResponse(nodePoolService.getProperties());
```
Pasichniuk: **Web layer reaching into `configuration/` directly.** Three sites do this:

- here, via `nodePoolService.getProperties()`
- `web/mapper/NodePoolDtoMapper.java:18` — `toListResponse(NodePoolProperties properties)`
- `web/mapper/DeploymentDtoMapper.java:111` — `@Autowired private NodePoolProperties nodePoolProperties`

Per `web/CLAUDE.md` the web layer "may call `service/`" — going through `NodePoolProperties` skips that boundary. Worse, `NodePoolService.getProperties()` returns the **raw mutable** properties bean — a caller could invoke setters on the live configuration object.

Two cleaner alternatives:
- Move `resolveNodePoolName(String id)` onto `NodePoolService` and inject the service into `DeploymentDtoMapper`.
- Have `NodePoolDtoMapper.toListResponse(...)` take a `List<PoolConfig>` and let this controller pass `nodePoolService.getNodePools()`.

Either way, drop `NodePoolService.getProperties()` — it has one purpose and that purpose is the leak.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -69,6 +69,7 @@ public class DeploymentService {
     private final DisposableResourceManager disposableResourceManager;
     private final HistoryService historyService;
     private final NodePoolProperties nodePoolProperties;
+    private final com.epam.aidial.deployment.manager.service.nodepool.NodePoolService nodePoolService;
```
Pasichniuk: Fully-qualified type name with no name collision — add a normal import statement at the top of the file instead. Same pattern appears in:
- `src/test/.../web/controller/none/DeploymentControllerTest.java` (`@MockitoBean private com.epam.aidial...NodePoolProperties`)
- `src/test/.../web/controller/none/internal/DeploymentInternalControllerTest.java` (same)

Cleanup nit but jarring to read.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -393,10 +402,22 @@ private static <T> Supplier<EntityNotFoundException> notFound(String what, T id)
...
+            throw new IllegalArgumentException("Node pool id '%s' is not configured".formatted(nodePoolId));
+        }
+    }
+
+    /**
+     * FR-018: when the create payload leaves {@code nodePoolId} null, resolve via the cascade and
+     * stamp the result onto the request so it persists onto the record. The duplicate flow opts
+     * out of this entirely by calling the worker with {@code applyNodePoolDefaultIfEmpty=false}.
+     */
+    private void applyCreateTimeNodePoolCascade(CreateDeployment request) {
```
Pasichniuk, line 417: **Cascade end-to-end is unit-tested at the helper level only.** `NodePoolServiceTest` covers `resolveForCreate(...)` thoroughly, but nothing exercises this method directly — i.e. that the resolved value is actually **stamped onto the request** before persistence, that the public/private overload split routes correctly (public passes `true`, `duplicateDeployment` passes `false`), and that FR-020 holds (duplicate bypasses the cascade).

Also no direct test for FR-013/FR-014 — that a non-null but unknown `nodePoolId` rejects with a 400 via `validateNodePoolId`.

The commit message for `381e40a` openly defers this ("a Spring-context integration test would require a test profile with NODE_POOLS populated and is deferred"). For a P1+P2 user story whose entire value proposition is *"the user sees the stamped value in the create response,"* at least one focused unit test of `DeploymentService` with a mocked `NodePoolService` would be cheap insurance against future refactors of the overload boundary.

src/main/java/com/epam/aidial/deployment/manager/web/dto/nodepool/NodePoolListResponseDto.java
```diff
@@ -0,0 +1,12 @@
+package com.epam.aidial.deployment.manager.web.dto.nodepool;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import java.util.List;
+
+@Schema(description = "Listing of configured node pools")
+public record NodePoolListResponseDto(
+        @Schema(description = "Configured pool catalogue")
+        List<NodePoolDto> pools
```
Pasichniuk, line 10: Missing `requiredMode = Schema.RequiredMode.REQUIRED`. FR-017 says:

> *The `pools` array MUST always be present in the response (serialised as `[]` when no pools are configured) so FE consumers can rely on its presence without null-checking.*

`NodePoolDto.id` / `name` correctly carry `requiredMode = REQUIRED`; mirror that on `pools` so the generated OpenAPI matches the contract.

```suggestion
        @Schema(description = "Configured pool catalogue", requiredMode = Schema.RequiredMode.REQUIRED)
        List<NodePoolDto> pools
```

src/test/java/com/epam/aidial/deployment/manager/service/config/imports/DeploymentImporterTest.java
```diff
@@ -127,6 +130,55 @@ void importDeployments_exists_overwrite_updatesDeployment() {
...
+        when(nodePoolService.resolveForCreate(createRequest)).thenReturn("target-default-pool");
+
+        deploymentImporter.importDeployments(config, ConflictResolutionPolicy.OVERWRITE);
+
+        verify(deploymentService).updateDeployment(eq(DEPLOYMENT_ID), eq(createRequest));
+        org.assertj.core.api.Assertions.assertThat(createRequest.getNodePoolId()).isEqualTo("target-default-pool");
+    }
+
+    @Test
+    void importDeployments_exists_overwrite_preservesExplicitNodePoolId() {
```
Pasichniuk, line 158: This test guards a code path the real import flow doesn't reach. `applyCreateTimeNodePoolCascade` only runs when `request.getNodePoolId() == null`, and for OVERWRITE imports the export mixin strips `nodePoolId` on read — so in production `createRequest.getNodePoolId()` is guaranteed null at this point. The only way the explicit-id branch is exercised is the way this test does it: by hand-constructing a `CreateDeployment` with `nodePoolId` set, bypassing the mapper.

Fine as a defensive guard, but worth noting because if someone later "fixes" `DeploymentMapper.toCreateDeployment(Deployment)` to pass through `nodePoolId` from the domain object (it currently doesn't), behaviour changes silently — the cascade quietly stops running on imports that happen to carry an id. A comment on the production code (`DeploymentImporter.importOne` OVERWRITE branch) explaining *why* the null-check is there would make that future change visible at the point of impact.

src/main/java/com/epam/aidial/deployment/manager/configuration/NodePoolProperties.java
```diff
@@ -1,95 +1,67 @@
...
+    private String defaultModelPoolId;
+
+    public Optional<PoolConfig> findById(String id) {
+        if (CollectionUtils.isEmpty(pools)) {
             return Optional.empty();
         }
-        return nodePools.stream()
-                .filter(pool -> pool.getName().equals(name))
+        return pools.stream()
+                .filter(pool -> pool.getId().equals(id))
```
Pasichniuk: `pool.getId().equals(id)` NPEs if `pool.getId()` is null. Validation guarantees non-blank ids at runtime (the `@NotBlank` check in `NodePoolConfiguration` rejects them at startup), so this is safe today. But `Objects.equals(pool.getId(), id)` costs nothing and survives someone constructing a `PoolConfig` directly in a test that bypasses validation.

```suggestion
                .filter(pool -> Objects.equals(pool.getId(), id))
```

src/main/java/com/epam/aidial/deployment/manager/service/manifest/PoolPrimitivesConverter.java
```diff
@@ -0,0 +1,49 @@
...
+    public <T> T convertAffinity(@Nullable Affinity source, Class<T> targetType) {
+        if (source == null) {
+            return null;
+        }
+        return mapper.convertValue(source, targetType);
+    }
+
+    @Nullable
+    public <T> List<T> convertTolerations(@Nullable List<Toleration> source, Class<T> targetElementType) {
+        if (CollectionUtils.isEmpty(source)) {
```
Pasichniuk, line 42: Returning `null` for empty input is asymmetric: `convertAffinity` does null→null (intuitive), but `convertTolerations` does empty→null (less so). The callers (`NimManifestGenerator`, `InferenceManifestGenerator`) null-check the result before merging, so behaviour is correct — just non-obvious.

Either return `List.of()` for the empty case and let the existing `CollectionUtils.isNotEmpty(convertedTolerations)` check at the call sites skip the merge, or add a one-line javadoc noting the empty-collapses-to-null contract.

src/main/java/com/epam/aidial/deployment/manager/service/nodepool/WorkloadClassifier.java
```diff
@@ -0,0 +1,16 @@
...
+import com.epam.aidial.deployment.manager.model.deployment.CreateDeployment;
+import com.epam.aidial.deployment.manager.model.deployment.CreateInferenceDeployment;
+import com.epam.aidial.deployment.manager.model.deployment.CreateNimDeployment;
+
+public final class WorkloadClassifier {
+
+    private WorkloadClassifier() {
+    }
+
+    public static boolean isModelWorkload(CreateDeployment deployment) {
```
Pasichniuk, line 19: **Closed enumeration with no safety net for future model deployment types.** FR-016 says "new deployment types added in the future MUST be classified in code as part of their introduction" — but nothing enforces that. A future model deployment subclass added without touching this file silently falls through to `false` and gets `NODE_POOL_DEFAULT` instead of `NODE_POOL_DEFAULT_MODEL`. The bug surfaces only as "model workloads land on the wrong pool," which an operator might attribute to misconfiguration rather than a missing classifier branch.

Two options that turn this into a compile-time check:
- Exhaustive `switch` on `CreateDeployment` (sealed type) so the compiler flags missing branches.
- Throw `UnsupportedOperationException` for unknown subclasses, forcing a code change when a new subtype lands.

Low priority — just a future foot-gun, not a bug today.

Pasichniuk (changes requested):

src/main/java/com/epam/aidial/deployment/manager/web/controller/NodePoolController.java
oleksii-donets: Fixed in 15a7a769. Dropped `NodePoolService.getProperties()` entirely and added `NodePoolService.findById(String)` for the one lookup callers actually need. `NodePoolDtoMapper.toListResponse` now takes `List<PoolConfig>` (controller passes `nodePoolService.getNodePools()`), and `DeploymentDtoMapper` injects `NodePoolService` instead of the raw properties bean. The two `@MockitoBean` declarations in the controller tests were swapped accordingly.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
oleksii-donets: Fixed in 15a7a769 — promoted to a normal import. (The two test sites referenced the same FQCN as a side effect of the previous `getProperties()` leak; they're gone now since those tests mock `NodePoolService` directly.)

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
oleksii-donets: Added a focused `DeploymentServiceTest` in 15a7a769 (pure Mockito, no Spring context). Four cases:

- `createDeployment_publicEntry_nullPoolId_stampsResolvedValue` — public entry with `nodePoolId == null` invokes `NodePoolService.resolveForCreate(req)` and the resolved id is stamped onto the request before persistence.
- `createDeployment_publicEntry_explicitPoolId_preserved` — explicit non-null id is preserved verbatim; cascade is never consulted.
- `createDeployment_publicEntry_unknownPoolId_throws` — unknown id triggers `IllegalArgumentException` from `validateNodePoolId` (covers FR-013/FR-014).
- `duplicateDeployment_skipsCascade` — verifies the private overload boundary by asserting `resolveForCreate` is never called via `duplicateDeployment` (FR-020).

src/main/java/com/epam/aidial/deployment/manager/web/dto/nodepool/NodePoolListResponseDto.java
oleksii-donets: Fixed in 15a7a769 — `pools` now carries `requiredMode = Schema.RequiredMode.REQUIRED`, matching the FR-017 guarantee.

src/test/java/com/epam/aidial/deployment/manager/service/config/imports/DeploymentImporterTest.java
oleksii-donets: Good catch — extended the comment in `DeploymentImporter.importOne` OVERWRITE branch (15a7a769) to call out that the guard is defensive: `DeploymentMapper.toCreateDeployment(Deployment)` does not currently pass `nodePoolId` through, so this branch is unreachable in production; the explicit null-check exists so that a future mapper change preserves FR-021 instead of silently disabling the cascade. The test stays as a regression net.

src/main/java/com/epam/aidial/deployment/manager/configuration/NodePoolProperties.java
oleksii-donets: Fixed in 15a7a769 — `Objects.equals(pool.getId(), id)`.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/PoolPrimitivesConverter.java
oleksii-donets: Fixed in 15a7a769. `convertTolerations` now returns `List.of()` for null/empty input, so both methods follow the "no input → harmless empty result" pattern. Dropped the `@Nullable` from the return type. Existing call sites already used `CollectionUtils.isNotEmpty(...)`, so no caller changes were needed; the two `PoolPrimitivesConverterTest` cases were updated to assert `isEmpty()`.

src/main/java/com/epam/aidial/deployment/manager/service/nodepool/WorkloadClassifier.java
oleksii-donets: Fixed in 15a7a769 — `isModelWorkload` is now a switch with an explicit branch per current `CreateDeployment` subclass and a `default` that throws `UnsupportedOperationException("Unknown CreateDeployment subtype, WorkloadClassifier must be updated: ...")`. Going with the runtime-exception variant rather than sealing `CreateDeployment` to keep the change local; the throw surfaces missed classifications immediately (the cascade tests exercise this code path) so FR-016 is enforced when a new subtype lands.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk:
Please do not add (node-pools) near feat or other tags - it breaks generation of release notes

oleksii-donets:
> Please do not add (node-pools) near feat or other tags - it breaks generation of release notes

@Pasichniuk conventional commit spec does not forbid it. Shouldn't we modify the generation of release notes instead?

Pasichniuk:
> > Please do not add (node-pools) near feat or other tags - it breaks generation of release notes
> 
> @Pasichniuk conventional commit spec does not forbid it. Shouldn't we modify the generation of release notes instead?

It will affect many projects, not just DM. You can request this improvement, if you want.

specs/README.md
```diff
@@ -32,6 +32,7 @@ Numbered specs created via `/speckit.specify`. `Status` reflects the value in ea
 | [016-node-pool-selector](016-node-pool-selector/spec.md) | Implemented | deployments | Node-pool selector API + per-deployment node pool assignment |
 | [017-stop-image-build](017-stop-image-build/spec.md) | Implemented | image-builds | Stop in-flight image builds from the UI/API |
 | [018-api-key-via-core-userinfo](018-api-key-via-core-userinfo/spec.md) | Implemented | security | Accept `Api-Key` header in oidc mode; validate via DIAL Core `/v1/user/info` |
+| [018-explicit-pool-scheduling](018-explicit-pool-scheduling/spec.md) | Implemented | deployments | Replace derived label-key + capacity-numbers pool config with explicit nodeSelector/affinity/tolerations; create-time defaults cascade for new deployments |
```
Pasichniuk: Please change feature number to 019

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #327, opened 2026-05-20
feat: revision rollback

### LINKED ISSUE #286: Implement rollback functionality
### Name and Version

ai-dial-admin-deployment-manager-backend [0.15.0]

Every DM write — to a deployment, an image definition, or the global image-build domain whitelist — is captured as an audit revision, but the audit history is **read-only**. When an administrator makes a bad change (a typo in an env var, a too-low replica count, the wrong source URL, an accidentally-removed whitelist entry), the only way back today is to open the audit snapshot and manually re-type every field through the regular update form.

That manual replay is:

- **Slow** — large entities (deployments with many env vars, whitelists with dozens of domains) take minutes to copy by hand.
- **Error-prone** — it is easy to miss a field or mistype a value, which turns "undo the last change" into "introduce a second bad change".
- **Audit-noisy** — the recovery shows up as a fresh manual edit, not as a deliberate revert, so the audit trail no longer cleanly reflects operator intent.
- **A blocker for safe experimentation** — administrators avoid otherwise-useful config changes because they cannot cheaply undo them.

Add a per-resource **rollback** action for the three audited write-heavy resources — deployments, image definitions, and the global image-build domain whitelist — that takes a target audit revision identifier and restores the resource's persisted configuration to its state at that revision in **one call**.

Key properties:

- **Scoped to one resource per call** — rolling back a deployment does not touch the image definition it references, and vice versa. No silent cascades.
- **Lifecycle-safe for deployments** — rollback is accepted only while the deployment is inactive (`NOT_DEPLOYED` or `STOPPED`); the operator flow for a live workload becomes `undeploy` → `rollback` → `deploy`. Running pods are never silently mutated.
- **Build-safe for image definitions** — rollback is accepted only when no usable build has been produced (`NOT_BUILT`, `BUILD_FAILED`, `BUILD_STOPPED`). A `BUILDING` or `BUILD_SUCCESSFUL` image definition may already be referenced by deployments, so its source/version is never silently shifted under them.
- **Audit-clean** — every state-changing rollback is recorded as a regular update activity attributed to the calling administrator, so the rollback itself is auditable and can be rolled back.
- **Same authorization as other writes** — `@FullAdminOnly`; read-only callers get HTTP 403.

### Demo scenarios

**1. Recover a deployment from a bad env-var change**

An admin updates a running deployment to fix a Helm-style typo and accidentally changes `LOG_LEVEL=info` to `LOG_LEVEL=debug` along with several unrelated env vars. The workload starts churning gigabytes of debug logs in production.

- **Today**: open the audit history, scroll to the previous snapshot, hand-copy every env var (15+ entries) back into the update form, hope no typo on the way back.
- **With rollback**: `undeploy` → `rollback` to the previous revision (one click) → `deploy`. Total time: seconds instead of minutes; zero risk of fat-fingering the recovery itself.

**Pro for the end user**: removes the largest source of operator toil and the largest blast-radius source of follow-on mistakes — recovery from a bad deployment edit is the most frequent "undo" scenario in DM.

**2. Revert an image definition after a failed build**

An admin points an MCP image definition at a new git ref to try out a fix; the build fails because the new ref pulls in a dependency that needs a domain not on the image definition's `allowedDomains` list. The admin wants to abandon the experiment and get back to the previous known-good source + metadata before triaging.

- **Today**: open the audit snapshot for the image definition, manually re-enter the git URL, ref, dockerfile path, allowed-domains list, builder selection — exactly the fields that are easiest to mis-copy.
- **With rollback**: rollback to the previous revision in one call. `buildStatus` is reset to `NOT_BUILT` automatically, so the next build attempt starts from a clean slate.

**Pro for the end user**: makes it cheap to **try** image-definition changes. Administrators stop avoiding source/version experiments because the cost of "undo" is now zero.

**3. Restore the global build-domain whitelist after an accidental deletion**

A platform admin is cleaning up the global image-build domain whitelist and removes what they think is an unused entry. Two hours later a scheduled rebuild of an unrelated image fails because that entry was the only thing allowing a transitive dependency mirror. Multiple build pipelines are now red.

- **Today**: read the whitelist update endpoint's replace-semantics carefully, manually re-type the full list of entries from the audit snapshot.
- **With rollback**: rollback the whitelist to the revision just before the cleanup. The current entries are replaced atomically with the snapshot's entries; if the whitelist is *already* in the target state (someone beat the operator to it), the rollback is a no-op and creates no spurious revision.

**Pro for the end user**: turns whitelist editing into a low-risk operation. The cost of being wrong drops from \"page on-call and restore by hand\" to \"click rollback\".

**4. Safe rollback of a rollback (audit-friendly experimentation)**

An admin rolls a deployment back to an earlier revision to investigate a regression, confirms the regression reproduces, and now wants to return to the latest configuration.

- **With rollback**: every rollback is itself recorded as a revision in the audit feed. The admin simply rolls the deployment forward to the revision that captured the latest state, with the same one-click flow.

**Pro for the end user**: rollback is bidirectional in practice. Administrators can use it as an investigation tool — bisecting through historical configurations to localize a regression — without polluting the audit trail or losing the ability to return to where they started.

- **Status quo (manual replay from snapshot endpoints)**: keeps the audit feed read-only and pushes recovery cost onto the operator. Rejected for the reasons in *What is the problem this feature will solve?*.
- **Cascading rollback (rolling back a deployment also rolls back the image definition it references, or vice versa)**: rejected — invisible cross-resource state changes are exactly the kind of "rollback surprise" this feature is trying to prevent. Each rollback is scoped to one resource.
- **Bulk \"rollback everything to revision R\"**: out of scope for the first cut. May be revisited if demand emerges.
- **Restoring deleted resources**: out of scope — the per-id rollback endpoint addresses currently-existing resources only. Forensic viewing of deleted resources remains available via the existing snapshot endpoints.

Pasichniuk:
### Applicable issues

- fixes #286 

### Description of changes

  - Adds a per-resource rollback action that restores deployments, image definitions, and the global image-build domain whitelist to a chosen past audit revision in one call, replacing today's manual field-by-field replay from the audit history.
  - Each rollback targets a single resource — there is no cascade between deployments and image definitions.
  - Deployment rollback is allowed only when the deployment is inactive (`NOT_DEPLOYED` or `STOPPED`); the operator flow for a live deployment becomes `undeploy` → `rollback` → `deploy`, so no running pods are silently mutated. Sensitive env-var values are reset to null on rollback and must be re-supplied before the next `deploy`.
  - Image-definition rollback is allowed only when the image has not produced a usable build (`NOT_BUILT`, `BUILD_FAILED`, `BUILD_STOPPED`); `BUILDING` and `BUILD_SUCCESSFUL` are rejected to avoid silently shifting the source/version of an image other deployments may already reference. If any field changes, build status resets to `NOT_BUILT`.
  - Global-whitelist rollback fully replaces the current entries with the snapshot's entries; an identical-state rollback is a no-op and does not create a redundant revision. Deployment and image-definition rollbacks always persist and record a new audit revision.
  - Returns the rolled-back resource's current representation, records the action as a regular update activity attributed to the caller, and inherits the existing `@FullAdminOnly` authorization — read-only callers are rejected with HTTP 403. Functional-test coverage is added for all three resources (H2, Postgres, SQL Server) and the Postman collection / capability specs (auditing, deployments, image-definitions, domain-whitelist) are updated accordingly.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/ImageDefinitionService.java
```diff
@@ -140,6 +140,33 @@ public ImageDefinition updateImageDefinition(UUID id, ImageDefinition updatedIma
...
+                .filter(other -> !other.getId().equals(id))
+                .ifPresent(other -> {
+                    throw new IllegalArgumentException(
+                            "Cannot roll back image definition '%s': name='%s' and version='%s' would collide with existing image definition '%s'."
+                                    .formatted(id, snapshot.getName(), snapshot.getVersion(), other.getId()));
+                });
+
+        return imageDefinitionRepository.updateImageDefinition(id, snapshot);
+    }
+
```
oleksii-donets, line 170: 🟠 **Spec gap vs. FR-003(a): image-def rollback does not re-run current validation.**

`specs/020-revision-rollback/spec.md` FR-003(a) explicitly mandates that image-definition rollback (and **only** image-definition rollback — deployments/whitelist are intentionally exempt per FR-003(e)) rejects with HTTP 400 when the snapshot would fail current validation rules.

This method only enforces the `BUILDING`/`BUILD_SUCCESSFUL` guard and `(name, version)` uniqueness. Calling `imageDefinitionRepository.updateImageDefinition(id, snapshot)` directly bypasses the `@Valid`-driven validators that protect the regular `PUT` endpoint (e.g. `DomainListValidator` on `allowedDomains`, image-builder presence, etc.), so a snapshot whose `allowedDomains` is no longer valid (or whose `imageBuilder` has been removed since) will round-trip into the row and only surface on the next regular update.

Two clean options:
- **Tighten the code**: invoke a validator inside `rollback` before calling the repo (mirror the `@Valid` chain).
- **Loosen the spec**: drop FR-003(a) for image-defs and accept the same "snapshot applies verbatim, surfaces on next write" asymmetry that deployments/whitelist have.

The new test scenarios cover the uniqueness path but not validation, so this gap isn't caught by CI today.

src/main/java/com/epam/aidial/deployment/manager/service/GlobalDomainWhitelistService.java
```diff
@@ -43,4 +44,15 @@ public List<String> getDomainWhitelistSnapshot(Integer revision) {
                 .orElseThrow(() -> new EntityNotFoundException(
                         "Unable to find domain whitelist at revision " + revision));
     }
+
+    @Transactional
+    public List<String> rollback(Integer revision) {
+        historyService.getRevisionById(revision);
+        var snapshot = getDomainWhitelistSnapshot(revision);
```
oleksii-donets, line 52: 🟡 **Rollback 404s when the current whitelist row does not exist.**

`repository.getAllowedDomains()` throws `GlobalDomainWhitelistNotFoundException` if no whitelist singleton row exists in the DB. So if a past revision has a snapshot but no current row (fresh DB, singleton was removed, etc.), this rollback fails with 404 — even though a valid snapshot exists and could be restored.

The newly-added `GlobalDomainWhitelistNotFoundException` mapping in `DefaultExceptionHandler` makes this a clean 404 rather than 500 (presumably why it was added in this PR), but if the spec's intent is "rollback restores the singleton when a snapshot exists", the cleaner code is:

```java
var current = repository.findAllowedDomainsOrEmpty(); // new helper, or use setAllowedDomainsOrCreate semantics
if (CollectionUtils.isEqualCollection(current, snapshot)) {
    return current;
}
repository.setAllowedDomainsOrCreate(snapshot);
return snapshot;
```

If the current 404 behaviour is deliberate, please add a scenario in `specs/domain-whitelist/spec.md` documenting it and a functional test asserting it, so the contract is explicit.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -159,6 +162,60 @@ private Deployment createDeployment(CreateDeployment request, boolean applyNodeP
...
+            if (env instanceof SensitiveFileEnvVar) {
+                sensitiveFile.put(env.getName(), null);
+            } else if (env instanceof SensitiveEnvVar) {
+                sensitive.put(env.getName(), null);
+            } else if (env instanceof SimpleEnvVar) {
+                nonSensitive.put(env.getName(), env.getValue());
+            }
+        }
+        return new EnvPartition(sensitive, nonSensitive, sensitiveFile);
+    }
```
oleksii-donets, line 222: 🟡 **`existing.getEnvs()` here has not been through `resolveSecrets` — worth a comment.**

`updateDeployment` (line 184) calls `deploymentManager.resolveSecrets(existingDeployment)` before any env handling. The rollback path skips that and hands the raw repository result straight to `cleanupSecrets`. It works correctly because `cleanupSecrets` only reads `k8sSecretName` off the `SensitiveEnvVar`, and that field is persisted in the DB — but the asymmetry will surprise the next reader.

A one-line comment on `reconcileEnvSecrets` saying e.g. *"we deliberately skip `resolveSecrets` here — `cleanupSecrets` keys off the persisted `k8sSecretName` only, no values needed"* would prevent someone "fixing" it later.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -159,6 +162,60 @@ private Deployment createDeployment(CreateDeployment request, boolean applyNodeP
         return savedDeployment;
     }
 
+    @Transactional
+    public Deployment rollback(String id, Integer revision) {
+        historyService.getRevisionById(revision);
```
oleksii-donets, line 168: 🟢 The return value of `historyService.getRevisionById(revision)` is discarded — it's called purely for its 404 side effect. Same pattern in `ImageDefinitionService.rollback` and `GlobalDomainWhitelistService.rollback`.

A one-line comment (e.g. `// 404 precheck: throws if the revision id is unknown`) would make intent obvious to anyone reading this for the first time.

src/main/java/com/epam/aidial/deployment/manager/web/controller/DeploymentController.java
```diff
@@ -96,6 +98,23 @@ public List<DeploymentDto> getAllDeploymentsAtRevision(@PathVariable Integer rev
...
+            description = "Restores the deployment's stored configuration to its snapshot at the given audit revision. "
+                    + "Allowed only when the deployment is in NOT_DEPLOYED or STOPPED. "
+                    + "Does not modify live Kubernetes state — the operator must trigger deploy afterwards to apply.")
+    @ApiResponse(responseCode = "200", description = "Rollback applied or identical-state no-op")
+    @ApiResponse(responseCode = "400", description = "Active-state, validation, or missing-reference rejection")
+    @ApiResponse(responseCode = "403", description = "Read-only role")
+    @ApiResponse(responseCode = "404", description = "Deployment or revision not found")
+    public DeploymentDto rollbackDeployment(@PathVariable String id, @PathVariable Integer revision) {
+        var rolledBack = deploymentService.rollback(id, revision);
+        return dtoMapper.toDeploymentDto(rolledBack);
```
oleksii-donets, line 115: 🟢 **Logging inconsistency across the three new rollback endpoints.**

`GlobalDomainWhitelistController.rollbackDomainWhitelistForImageBuild` logs start + success at INFO with the revision id. This one and `ImageDefinitionController.rollbackImageDefinition` log nothing.

For an operator-driven recovery action you probably want all three to log start + success with the deployment / image-def id and revision id, mirroring how update/delete endpoints log today. Pick one convention and apply it across the three.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
```diff
@@ -0,0 +1,427 @@
...
+                .orElse(null);
+    }
+
+    private Integer createRevisionFor(String resourceId) {
+        return createRevisionForString(resourceId);
+    }
+
+    private Integer createRevisionForString(String resourceId) {
+        return createRevisionByPredicate(a -> a.getActivityType() == ActivityType.Create
+                && resourceId.equals(a.getResourceId()));
```
oleksii-donets, line 408: 🟢 `createRevisionFor(String)` just delegates to `createRevisionForString`, which does exactly what `createRevisionFor` would do inline. The indirection serves no purpose — inline one of them.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
```diff
@@ -0,0 +1,427 @@
...
+        // Simulate an active state by directly mutating status through the repository.
+        Deployment loaded = deploymentRepository.getById(created.getId()).orElseThrow();
+        loaded.setStatus(DeploymentStatus.RUNNING);
+        deploymentRepository.update(created.getId(), loaded);
+
+        assertThatThrownBy(() -> deploymentService.rollback(created.getId(), createRevision))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("active state");
+    }
+
```
oleksii-donets, line 146: 🟢 Simulating the `RUNNING` state by directly writing through the repo is fine for testing the service-level guard, but no production flow reaches `RUNNING` via `repository.update`. If someone later adds a transition guard at the repository layer (e.g. forbid arbitrary status writes), this test will fail for the wrong reason. Worth a comment noting it's a deliberate test backdoor.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
```diff
@@ -0,0 +1,427 @@
...
+        ImageDefinition anchorCreated = imageDefinitionService.createImageDefinition(anchorImageDef);
+        Integer earlierRevision = createRevisionForUuid(anchorCreated.getId());
+
+        // Now create the subject deployment — it exists from a later revision.
+        CreateDeployment request = createInterceptorDeploymentWithoutSecrets("rollback-test-5");
+        deploymentService.createDeployment(request);
+
+        assertThatThrownBy(() -> deploymentService.rollback("rollback-test-5", earlierRevision))
+                .isInstanceOf(EntityNotFoundException.class);
+    }
```
oleksii-donets, line 193: 🟢 This case asserts only `EntityNotFoundException`, but both "revision id is unknown" (R4) and "revision exists but predates the resource" (this test, R5) currently map to the same exception type via Envers. Adding `.hasMessageContaining("revision")` or `.hasMessageContaining("Deployment")` (whichever message wins from `entitySnapshotAtRevision`) would distinguish the two — otherwise a future change that breaks R5 but leaves R4 working could pass both tests.

src/main/java/com/epam/aidial/deployment/manager/service/ImageDefinitionService.java
Pasichniuk: Updated the spec

src/main/java/com/epam/aidial/deployment/manager/service/GlobalDomainWhitelistService.java
Pasichniuk: Fixed

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
Pasichniuk: Added

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
Pasichniuk: Added

src/main/java/com/epam/aidial/deployment/manager/web/controller/DeploymentController.java
Pasichniuk: Intentional. We log all updates for global whitelist, that's why rollback is also logged. Not needed for image definitions and deployments because they are updated more frequently, therefore producing a lot of noise in logs.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
Pasichniuk: Done

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
Pasichniuk: Done

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
Pasichniuk: Done

ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

PR #334, opened 2026-05-21
feat: auto-detect HF text-classification and chain KServe transformer

### LINKED ISSUE #87: [Model Servings] Support text classification models
### Name and Version

Acceptance Criteria

_No response_

oleksii-donets:
### Applicable issues

- fixes #87

### Description of changes

Implements feature [021-inference-task-transformer](specs/021-inference-task-transformer/spec.md): auto-detected HuggingFace text-classification deployments are now emitted as a chained KServe `InferenceService` (predictor + transformer), with zero operator-facing task/id2label fields.

**Detection (deploy-time, no schema changes)**
- New `InferenceTaskDetector` reads HF Hub metadata for the deployment's model at manifest-generation time and computes `task` + `id2label`. No new columns, no migration, no DTO fields — detection is fetched fresh on every deploy.
- `pipeline_tag = "text-classification"` → chained manifest; everything else → predictor-only (preserves backward compatibility for existing non-classification deployments).
- Typed exception hierarchy maps every failure to its spec'd HTTP status: `ModelMetadataMissingException` / `ModelMetadataUnusableException` / `ModelNotFoundException` → 400, `HuggingFaceUpstreamException` → 502, `MissingTransformerImageException` → 500. `DefaultExceptionHandler` and the deploy endpoint's `@ApiResponse` set are updated accordingly.

**Manifest (chained predictor + transformer)**
- `TextClassificationTransformerSection` generates the transformer block; predictor args are auto-injected with `--return_raw_logits` and `--task=sequence_classification`, predictor block carries `protocolVersion: v2`.
- Transformer container resource requests/limits and image are bound from `AppProperties` as a raw Fabric8 `Container` (cluster-wide config via env vars; no per-deployment knob).
- Transformer config pins `config.json` to the model's commit SHA so label drift can't sneak in mid-deploy.
- Public URL of a chained deployment resolves to the transformer component.

**Argument-contract validation**
- Operator-supplied predictor `command`/`args` that violate the chained-component contract are rejected at the API boundary rather than failing silently at runtime.

**Spec hygiene**
- 021 `Status` → Implemented; auto-detect / unusable-metadata Requirements added to `specs/inference-deployments/spec.md`; chained predictor+transformer Requirement added to `specs/kubernetes-manifests/spec.md`; 021 row added to `specs/README.md`.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
✅ No vulnerabilities or license issues or OpenSSF Scorecard issues found.<h2>OpenSSF Scorecard</h2>
<table><tr><th>Package</th><th>Version</th><th>Score</th><th>Details</th></tr>
<tr><td> maven/org.apache.tomcat.embed:tomcat-embed-core </td><td>11.0.22</td>
      <td> Unknown</td><td>Unknown</td></tr>
</table><h2>Scanned Files</h2>
<ul><li>settings.gradle</li></ul>

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/huggingface/client/HuggingFaceClient.java
```diff
@@ -75,6 +78,79 @@ public ModelsPageResponse getModelsPage(ModelsRequest request, String pageUrl) {
...
+                .revision(resolvedRevision)
+                .filePath(CONFIG_JSON_PATH)
+                .build();
+        try (var body = downloadFile(fileRequest)) {
+            var result = jsonMapper.readValue(body.string(), ModelConfig.class);
+            log.debug("huggingface model config.json retrieved. Model: {}. Revision: {}", modelName, resolvedRevision);
+            return result;
+        } catch (HuggingFaceClientException e) {
+            throw e;
+        } catch (com.fasterxml.jackson.core.JsonProcessingException e) {
```
Pasichniuk: Redundant full path

src/main/java/com/epam/aidial/deployment/manager/service/detection/HuggingFaceUpstreamException.java
```diff
@@ -0,0 +1,13 @@
+package com.epam.aidial.deployment.manager.service.detection;
```
Pasichniuk: Let's keep all exceptions in 'com.epam.aidial.deployment.manager.exception' package

src/main/java/com/epam/aidial/deployment/manager/web/controller/DeploymentController.java
```diff
@@ -166,6 +167,16 @@ public void deleteDeployment(@PathVariable String id) {
...
     @FullAdminOnly
+    @Operation(summary = "Deploy a deployment by id")
+    @ApiResponses({
+        @ApiResponse(responseCode = "200", description = "Deployment scheduled for deploy"),
+        @ApiResponse(responseCode = "400", description = "HuggingFace model not found, unauthorized, has"
+                + " unusable metadata, or supplied predictor arg conflicts with the chained transformer contract."),
+        @ApiResponse(responseCode = "500", description = "Required transformer-image configuration is not set;"
+                + " deploy aborted before cluster mutation."),
+        @ApiResponse(responseCode = "502", description = "HuggingFace Hub unreachable; deploy not started,"
+                + " safe to retry.")
```
Pasichniuk: There are many more possible api responses besides these. Either add all of them or remove ApiResponses block

CLAUDE.md
```diff
@@ -70,3 +70,10 @@ Where `<slug>` is the directory name under `specs/` that the feature belongs to
...
 3. If the field is `N/A — creates new capability <slug>`, scaffold `specs/<slug>/spec.md` from the house style (use `specs/api-conventions/spec.md` as a model) and add a row to `specs/README.md`.
+
+## Active Technologies
+- Java 21 (constitution §Tech Stack) + Spring Boot 3.5.10, Lombok via `io.freefair.lombok` 8.10, MapStruct 1.6.0, Fabric8 Kubernetes Client 7.5.2 (KServe CRD generated), Flyway 11.14.0, Hibernate Envers, OkHttp (transitive — used by existing `HuggingFaceClient`), Jackson 2.21.1 (021-inference-task-transformer)
+- H2 / PostgreSQL / SQL Server (multi-vendor; existing `inference_deployment` table). New columns added via Flyway migration V1.59 (next free version). (021-inference-task-transformer)
+
+## Recent Changes
+- 021-inference-task-transformer: Added Java 21 (constitution §Tech Stack) + Spring Boot 3.5.10, Lombok via `io.freefair.lombok` 8.10, MapStruct 1.6.0, Fabric8 Kubernetes Client 7.5.2 (KServe CRD generated), Flyway 11.14.0, Hibernate Envers, OkHttp (transitive — used by existing `HuggingFaceClient`), Jackson 2.21.1
```
Pasichniuk: Redundant change

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
```diff
@@ -122,9 +134,73 @@ public InferenceService serviceConfig(
...
+            if (arg.equals(TASK_ARG) || arg.startsWith(TASK_ARG + "=")) {
+                int eq = arg.indexOf('=');
+                String value = (eq < 0) ? "" : arg.substring(eq + 1);
+                if (!SEQUENCE_CLASSIFICATION_TASK.equals(value)) {
+                    throw new IllegalArgumentException(("Inference deployment '%s' is a text-classification model"
+                            + " and cannot override '--task' to '%s' — the chained transformer requires"
+                            + " '--task=%s'. Remove this arg or set it to the required value.")
+                            .formatted(name, value, SEQUENCE_CLASSIFICATION_TASK));
+                }
+            }
```
Pasichniuk: **Bug — space-form `--task <value>` is always rejected.**

When the operator supplies args in space form, e.g. `["--task", "sequence_classification"]`, the loop visits the bare `"--task"` token: `arg.equals(TASK_ARG)` is true, `eq = -1`, `value = ""`, and `"sequence_classification".equals("")` is false → throws.

Result: the **correct** value `sequence_classification` is rejected, and the error message reads `cannot override '--task' to ''` — misleading. The codebase clearly supports space-form predictor args (see `setModelNameIfNotSet` at line 245, which adds `[--model_name, modelName]` as separate entries; `isArgPresent` accepts both forms).

**Suggested fix:** when `arg.equals(TASK_ARG)`, look at the next list element for the value (or treat the bare flag as a parse error with a clearer message). Same applies to `--return_probabilities` space form on line 186.

**Test gap:** `InferenceManifestGeneratorTest` covers only equals-form (`--task=summarization`, `--task=sequence_classification`). Add explicit space-form cases.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
```diff
@@ -122,9 +134,73 @@ public InferenceService serviceConfig(
...
+        modelChain.data().setProtocolVersion(KSERVE_V2_PROTOCOL);
+
+        var predictorArgs = modelChain.get(InferenceMappers.MODEL_ARGS_FIELD).data();
+        rejectConflictingChainedArgs(name, predictorArgs);
+        if (!isArgPresent(RETURN_RAW_LOGITS_ARG, predictorArgs)) {
+            predictorArgs.add(RETURN_RAW_LOGITS_ARG);
+        }
+        if (!isArgPresent(TASK_ARG, predictorArgs)) {
+            predictorArgs.add(TASK_ARG + "=" + SEQUENCE_CLASSIFICATION_TASK);
+        }
```
Pasichniuk, line 174: **Bug / spec violation — operator can silently bypass `--return_raw_logits`.**

`isArgPresent` (line 248) matches both `--return_raw_logits` and `--return_raw_logits=*`. An operator supplying `--return_raw_logits=false` will:
- not be rejected by `rejectConflictingChainedArgs` (only `--return_probabilities` and `--task=<other>` are forbidden),
- not trigger auto-injection here (because `isArgPresent` reports the flag as already present),
- get a chained manifest whose predictor returns probabilities — silently breaking the transformer contract.

Spec FR-014 says: "the predictor's container args SHALL **always** include `--return_raw_logits` and `--task=sequence_classification` (auto-injected by the system). Operator-supplied `command`/`args` are merged on top — they do not replace these."

**Suggested fixes (pick one):**
- Add `--return_raw_logits=<falsey>` to the forbidden-args list (preferred — surfaces the conflict).
- Always append the flag and let the predictor CLI dedup / last-write-wins.
- Validate that any operator-supplied `--return_raw_logits=*` matches the chained contract.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/InferenceDeploymentManager.java
```diff
@@ -199,15 +208,42 @@ protected DeploymentStatus mapStatus(InferenceService service) {
...
+        var transformer = status.getComponents().get("transformer");
+        if (transformer == null) {
+            return false;
+        }
+        if (transformer.getUrl() != null) {
+            return true;
+        }
+        var address = transformer.getAddress();
+        return address != null && address.getUrl() != null;
+    }
```
Pasichniuk, line 255: **Weak transformer readiness check.** This only checks URL presence; there's no equivalent of the predictor's `ActiveModelState=LOADED && TransitionStatus=UPTODATE`. A transformer that had its URL set in a prior reconcile but then crashed would still report `RUNNING` for the parent deployment.

Spec FR-019 / SC-007 require "RUNNING only when **both** components are healthy". URL presence is a soft proxy for that. Consider checking whether the KServe `transformer` component status carries equivalent readiness/transition fields and gating on them. If KServe doesn't expose component-level transition status for transformer, leaving a TODO documenting the limitation would be useful for the next maintainer.

src/main/java/com/epam/aidial/deployment/manager/service/detection/InferenceTaskDetector.java
```diff
@@ -0,0 +1,190 @@
...
+     * @throws ModelMetadataUnusableException if {@code id2label} fails the structural contract
+     */
+    public InferenceTaskDetectionResult detect(HuggingFaceSource source) {
+        var modelName = source.modelName();
+        log.debug("Detecting inference task for HF model '{}'", modelName);
+
+        Model model = fetchModel(modelName);
+        // Pin config.json to the same revision the model API returned so the two calls
+        // can't observe different snapshots if the model is updated between them.
+        ModelConfig config = fetchConfig(modelName, model.getSha());
```
Pasichniuk, line 67: **`config.json` is always fetched even when `pipeline_tag` already settles the question.**

For non-classification models that correctly tag themselves (the common case), this doubles HF Hub API load on every deploy — `config.json` is only needed for the architecture fallback (NONE-tagged sequence-classification models) and for the `id2Label` map (when classification). A short-circuit on `pipeline_tag = text-classification` could skip the architecture fallback's config fetch.

The earlier commit message ("control flow flattened to a single linear pass") notes this was a deliberate simplification. Either keep this and add a one-line comment explaining the trade-off, or split into two paths.

Not blocking.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/TextClassificationTransformerSection.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private Containers toKserveContainer(String deploymentName, Container source) {
+        try {
+            var json = jsonMapper.writeValueAsString(source);
+            return jsonMapper.readValue(json, Containers.class);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException(
+                    "Failed to convert transformer container template for deployment '%s'".formatted(deploymentName), e);
+        }
+    }
```
Pasichniuk, line 131: **Fragile JSON round-trip between Fabric8 `Container` and generated KServe `Containers`.**

This serializes a Fabric8 `io.fabric8.kubernetes.api.model.Container` and deserializes into the KServe-generated `inferenceservicespec.transformer.Containers`. As long as the two types stay schema-compatible this works, but a future KServe CRD bump that renames or drops a field will be silently swallowed by Jackson's `@JsonIgnoreProperties(ignoreUnknown = true)` defaults — operators tuning that field in YAML will see no error and the manifest won't have what they configured.

Either (a) document the schema-coupling assumption in the class javadoc, or (b) build the `Containers` directly via Fabric8 builders. Option (b) is more code but eliminates the silent-drop risk.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
```diff
@@ -122,9 +134,73 @@ public InferenceService serviceConfig(
...
+
+        textClassificationTransformerSection.apply(service, name, id2Label);
```
Pasichniuk, line 176: **Order: predictor is mutated before transformer template is validated.**

By the time `textClassificationTransformerSection.apply(...)` throws `MissingTransformerImageException`, the `InferenceService` already has `protocolVersion=v2`, `--return_raw_logits`, and `--task=sequence_classification` mutated onto its predictor. Since `serviceConfig` returns the spec only on success and the caller discards on throw, this has no externally visible effect today — but it makes the method order-dependent in a non-obvious way.

Flipping the order (validate transformer template first, then mutate predictor) is safer and clearer.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
```diff
@@ -122,9 +134,73 @@ public InferenceService serviceConfig(
...
+     * </ul>
+     */
+    private void applyChainedTransformer(String name,
+                                         @Nullable Map<Integer, String> id2Label,
+                                         MappingChain<Model> modelChain,
+                                         InferenceService service) {
+        if (MapUtils.isEmpty(id2Label)) {
+            throw new IllegalStateException("Cannot emit chained transformer for deployment '%s': id2Label is empty"
+                    .formatted(name));
+        }
```
Pasichniuk: Defense-in-depth, but the detector's invariants guarantee `id2Label != null && !isEmpty()` whenever `task == TEXT_CLASSIFICATION`. If kept, `Objects.requireNonNull` with a precondition message is cleaner than `IllegalStateException` (which here ends up at the catch-all 500 handler rather than something more specific). Or just drop the check.

specs/021-inference-task-transformer/data-model.md
```diff
@@ -0,0 +1,105 @@
...
+| Import / export | No detected fields participate. Existing deployment fields round-trip as before. |
+
+## Configuration entity (unchanged from revision 1)
+
+### `TextClassificationTransformerProperties`
+
+Package: `com.epam.aidial.deployment.manager.configuration`
+
+```text
+@ConfigurationProperties("app.inference.text-classification-transformer")
```
Pasichniuk: **Doc drift.** This section still documents the original `@ConfigurationProperties("app.inference.text-classification-transformer")` + bespoke `TextClassificationTransformerProperties` design, but the shipped implementation binds under `app.text-classification-transformer-container-config` as a raw Fabric8 `Container` (per the third commit's refactor: "bind transformer container as raw K8s Container under AppProperties").

The revision-2 note at the top of this file only covers the persistence drop, not the property-binding refactor. Add a revision-3 note (or rewrite this section to match the shipped design) so readers don't look for a class that doesn't exist.

Pasichniuk (changes requested):
Great spec hygiene throughout: capability specs updated with `Implemented via 021-…` cross-refs, typed exception hierarchy with depth-based handler resolution, sha-pinning between `getModel` and `fetchModelConfig`, deterministic id2label ordering. The chained-transformer happy paths are well-covered.

The two predictor-arg bugs (1 & 2 below) are the most important — they have concrete operator-facing impact and the existing test suite doesn't exercise them.

specs/021-inference-task-transformer/data-model.md
```diff
@@ -0,0 +1,105 @@
+# Phase 1 — Data Model
```
Pasichniuk, line 1: These files can be dropped safely if you want: data-model.md, plan.md, quickstart.md, tasks.md and inference-deployment-api.md. They are useful for spec-kit only during feature implementation, but since it's already implemented - only spec.md and research.md are useful in future. Maybe it'll be worth documenting that in CLAUDE.md later in a separate PR.

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/huggingface/client/HuggingFaceClient.java
oleksii-donets: Done in c5e295ac (already in the previous push via the working-tree change): the FQN is replaced with an explicit `import com.fasterxml.jackson.core.JsonProcessingException;` and the catch clause now uses the short name.

src/main/java/com/epam/aidial/deployment/manager/service/detection/HuggingFaceUpstreamException.java
oleksii-donets: Done in c5e295ac. All 6 inference-detection / transformer exceptions (`InferenceTaskDetectionException`, `HuggingFaceUpstreamException`, `ModelNotFoundException`, `ModelMetadataMissingException`, `ModelMetadataUnusableException`, `MissingTransformerImageException`) are now under `com.epam.aidial.deployment.manager.exception`; importers and tests updated.

src/main/java/com/epam/aidial/deployment/manager/web/controller/DeploymentController.java
oleksii-donets: Removed in c5e295ac. The block was a partial subset and would have grown stale fast; errors already flow through `DefaultExceptionHandler` / `ErrorView` and are documented at the type level via the typed exception hierarchy plus `specs/021-inference-task-transformer/contracts/inference-deployment-api.md` and `specs/api-conventions/spec.md`.

CLAUDE.md
oleksii-donets: Removed in c5e295ac — the speckit-generated `## Active Technologies` / `## Recent Changes` blocks are dropped from `CLAUDE.md`.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
oleksii-donets: Already addressed in 3a203b43 (before c5e295ac) — `rejectConflictingChainedArgs` switched to an index-based loop that handles both forms: split (`["--task", "<value>"]` peeks at the next token) and fused (`--task=<value>`). Test coverage was added in the same commit: `testServiceConfig_chained_acceptsTaskSplitFormSequenceClassification` (line 518) and `testServiceConfig_chained_rejectsTaskSplitFormConflictingValue` (line 536) in `InferenceManifestGeneratorTest`.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
oleksii-donets: Fixed in c5e295ac — went with the preferred option: `rejectConflictingChainedArgs` now treats any `--return_raw_logits=<non-true>` as a conflict (case-insensitive). Bare `--return_raw_logits` and `--return_raw_logits=true` are still accepted. New tests in `InferenceManifestGeneratorTest`: `testServiceConfig_chained_rejectsReturnRawLogitsFalseOverride` and `testServiceConfig_chained_acceptsReturnRawLogitsTrueOverride`.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/InferenceDeploymentManager.java
oleksii-donets: Documented in c5e295ac. Added a javadoc on `isTransformerReady` calling out that KServe does not surface a per-component equivalent of the predictor's `ActiveModelState=LOADED && TransitionStatus=UPTODATE` pair; URL presence is the closest signal in `InferenceServiceStatus.components` today. The doc explicitly notes the crash-after-URL-set blind spot and points the next maintainer at richer per-component fields if KServe adds them.

src/main/java/com/epam/aidial/deployment/manager/service/detection/InferenceTaskDetector.java
oleksii-donets: Kept the linear pass and added the one-line trade-off comment in c5e295ac (`InferenceTaskDetector.detect`, just above the `fetchConfig` call). The deliberate simplification is preserved; a future split-path optimization is now flagged inline for whoever notices HF Hub QPS becoming a concern.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/TextClassificationTransformerSection.java
oleksii-donets: Went with option (a) in c5e295ac. `toKserveContainer` now has javadoc spelling out the structural-identity assumption between Fabric8's `Container` and the KServe-generated `Containers`, noting that a future CRD bump renaming/dropping a field would be silently swallowed by Jackson's default `ignoreUnknown`, and pointing at Fabric8 builders as the escape hatch if that ever materializes.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
oleksii-donets: Fixed in c5e295ac. `TextClassificationTransformerSection` now exposes `ensureConfigured()` which only validates the image config; `applyChainedTransformer` calls it at the top, before the `setProtocolVersion` / arg mutations. A missing transformer image now throws before the predictor is touched.

src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
oleksii-donets: Dropped in c5e295ac. The detector's invariant (`task == TEXT_CLASSIFICATION ⇔ id2Label is non-null and non-empty`) already enforces this; removed the `@Nullable` from the local `applyChainedTransformer` parameter so the contract is explicit at the type level, and deleted the `IllegalStateException` check (which mapped to 500 — worse than the catch-all NPE would have been anyway).

specs/021-inference-task-transformer/data-model.md
oleksii-donets: Updated in c5e295ac. Added a "Revision 3 (2026-05-22)" callout at the top of `data-model.md` and rewrote the `## Configuration entity` section to document the shipped binding: a raw Fabric8 `Container` on `AppProperties` under `app.text-classification-transformer-container-config`, with the deployment-specific overlays (`--model_name`, `--predictor_protocol`, `ID2LABEL`) applied by `TextClassificationTransformerSection.apply`.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #337, opened 2026-05-21
fix: early failure detection for builder-push containers

### LINKED ISSUE #294: [Deployments] If the build failed we're waiting 8 minutes
### Name and Version

[BE]0.16.0-dev.29

1. Create image and fail it

If the build failed we're not waiting

Image

### What do you see instead?

If the build failed we're waiting 8 minutes

### Additional information

_No response_

Pasichniuk:
### Applicable issues

- fixes #294 

### Description of changes

  - Wrapped buildctl-daemonless.sh in both the root and rootless builder containers with a /bin/sh -c invocation that touches /image-build/build-failed on non-zero exit (preserving the original exit code via "$@" so the runtime-appended --local context=... args still reach buildctl).
  - Added a sentinel check at the top of the push container's polling loop so it exits 1 immediately when the builder reports failure, instead of running out its full 120 × 2 s = 240 s retry budget waiting for a tarball that will never arrive.
  - Eliminates the ~8 min wait reported in #294: when the builder fails, the pod's containers now both terminate within one push-poll cycle (~2 s), the Job transitions to Failed, and the build status flips to BUILD_FAILED in seconds rather than minutes — no behavior change on the success path.
  - Side-change: increased image-build-stop-timeout-sec from 30s to 45s, since 30 is too low for dev env

### Checklist

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -118,7 +118,7 @@ app:
   image-name-format: ${IMAGE_NAME_FORMAT:app-%s}
   image-label: latest
   image-build-timeout-sec: 300
-  image-build-stop-timeout-sec: 30
+  image-build-stop-timeout-sec: 45
   image-build-logs-size-limit: 1000
   heartbeat-period-sec: 30
   max-error-log-lines: 20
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

PR #338, opened 2026-05-22
fix: align rollback revision lookup with snapshot semantics

Pasichniuk:
### Applicable issues

- related to #286 

### Description of changes

  - Made the three "rollback to a past revision" operations (deployments, image definitions, global image-build whitelist) behave the same way as the matching "view what this looked like at a past revision (snapshot)" endpoints.
  - Why: operators choose a revision out of the audit history and reasonably expect "if I can view the state at this revision, I can roll back to it." Previously the rollback path was stricter and would reject some revisions that the view path happily accepted — surprising and inconsistent.
  - Concrete effect: asking to roll back to a revision number that falls in a gap of the audit log (e.g. the revision belongs to a different resource than the one being rolled back) now succeeds and restores the most recent applicable earlier state, instead of returning a not-found error.
  - Tests and the related specs were updated alongside the code so the documented behaviour matches what now ships, and the cross-repo ai-dial-admin-backend rollback already works this way — so this also aligns the two services.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -164,9 +164,6 @@ private Deployment createDeployment(CreateDeployment request, boolean applyNodeP
 
     @Transactional
     public Deployment rollback(String id, Integer revision) {
-        // 404 precheck: throws if the revision id is unknown
-        historyService.getRevisionById(revision);
-
         var existing = deploymentRepository.getById(id).orElseThrow(notFound("Deployment", id));
```
oleksii-donets, line 175: Concern: this removes more than the PR description suggests.

`historyService.getRevisionById` queries the `revinfo` table, which gets a row for **every** audited transaction regardless of which entity was touched. A revision that "belongs to a different resource" (the PR's stated gap case) *is* in `revinfo` and was already passing this precheck — the new `shouldRollbackSuccessfully_whenTargetRevisionHasGap` test below would have passed against the old code too.

The revisions the precheck actually rejected are ones with no `revinfo` row at all (e.g. `999_999`, or any value past the current max). With the precheck gone, `auditReader.find(DeploymentEntity, id, 999_999)` leniently returns the latest entity snapshot — i.e. the deployment's current state — and the rollback proceeds. `reconcileEnvSecrets` then unconditionally nulls every sensitive env value (there's no `snapshot == current` short-circuit on this path), forcing the operator to re-supply them via `updateDeployment` before the next deploy. Previously the same call returned 404 and left the deployment untouched.

Was this trade-off intentional? If yes, please call out the env-secret side effect for out-of-range revisions in the spec. If not, a targeted `revision <= historyService.maxRevisionId()` (or equivalent) would close the typo footgun while still honouring real gap revisions — which, again, were never rejected to begin with.

src/main/java/com/epam/aidial/deployment/manager/service/ImageDefinitionService.java
```diff
@@ -142,9 +142,6 @@ public ImageDefinition updateImageDefinition(UUID id, ImageDefinition updatedIma
 
     @Transactional
     public ImageDefinition rollback(UUID id, Integer revision) {
-        // 404 precheck: throws if the revision id is unknown
-        historyService.getRevisionById(revision);
-
         var existing = imageDefinitionRepository.getImageDefinitionById(id)
```
oleksii-donets, line 153: Same root concern as `DeploymentService.rollback` — see the comment there. The image-definition path is safer (no env-secret reconciliation, and the name+version uniqueness check stays in place), but a typo'd / out-of-range revision id will now produce a `buildStatus = NOT_BUILT` audit revision on an otherwise-unchanged image definition because rollback always calls `updateImageDefinition`. Worth confirming that's intended.

src/main/java/com/epam/aidial/deployment/manager/service/GlobalDomainWhitelistService.java
```diff
@@ -47,8 +47,6 @@ public List<String> getDomainWhitelistSnapshot(Integer revision) {
 
     @Transactional
     public List<String> rollback(Integer revision) {
-        // 404 precheck: throws if the revision id is unknown
-        historyService.getRevisionById(revision);
         var snapshot = getDomainWhitelistSnapshot(revision);
```
oleksii-donets, line 59: For the whitelist this is the most defensible of the three removals — `rollback` already short-circuits when `CollectionUtils.isEqualCollection(current, snapshot)` (line 52), so a typo'd / out-of-range revision id resolves to current and returns without persisting. No regression here. Worth differentiating from the deployment case in the spec.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
```diff
@@ -163,12 +163,27 @@ void shouldRollbackSuccessfully_whenAlreadyMatchingTargetRevision() {
     }
 
     @Test
-    void shouldFailRollback_whenRevisionDoesNotExist() {
-        CreateDeployment request = createInterceptorDeploymentWithoutSecrets("rollback-test-4");
-        deploymentService.createDeployment(request);
+    void shouldRollbackSuccessfully_whenTargetRevisionHasGap() {
```
oleksii-donets, line 179: This test doesn't actually exercise the behaviour change in this PR.

`gapRevision` comes from `createRevisionForUuid(unrelatedCreated.getId())` — a revision id that exists in `revinfo`. The old `getRevisionById` precheck would have accepted it; `entitySnapshotAtRevision` would have leniently resolved it to the deployment's R1 state. The test passes with or without the precheck.

To lock down the new semantics (the case the old precheck *did* reject), please add a test with an out-of-range revision id (e.g. `Integer.MAX_VALUE` or `latestRevisionId() + 1000`) asserting whatever the intended new behaviour is — especially for the deployment path, where the env-secret reconciliation side effect matters. Mirror in the image-definition and whitelist tests.

specs/020-revision-rollback/spec.md
```diff
@@ -93,7 +93,7 @@ A platform administrator notices that a recent change to the global image-build
 ### Functional Requirements
 
 - **FR-001**: The system MUST expose a per-resource rollback operation for deployments, image definitions, and the global image-build domain whitelist that takes a target audit revision identifier and restores the resource's persisted configuration to its state at that revision.
-- **FR-002**: The system MUST reject the rollback with HTTP 404 when (a) the resource does not currently exist under the supplied identifier, (b) the supplied revision identifier does not exist, or (c) the resource did not exist at the supplied revision.
+- **FR-002**: The system MUST reject the rollback with HTTP 404 when (a) the resource does not currently exist under the supplied identifier, or (b) the resource did not exist at the supplied revision. The system MUST resolve the target revision to the resource's state as of the latest applicable revision ≤ the target — matching the snapshot endpoint's lenient point-in-time semantics. Consequently, a target revision id that has no row in `revinfo` but is between revisions at which the resource was modified does not by itself trigger 404; rejection happens only when no restorable state exists at-or-before that revision.
```
oleksii-donets: Wording is slightly off relative to what Envers actually does.

"A target revision id that has no row in `revinfo` but is between revisions at which the resource was modified" conflates two distinct cases:

- **In `revinfo`, but no modification of this resource happens to fall on it.** The gap case. The old precheck already let this through (`revinfo` has a row), so no behaviour change here.
- **Not in `revinfo` at all** — typo'd / out-of-range revision id. This is what the precheck rejected, and what this PR now allows.

Suggest re-anchoring FR-002 on what `auditReader.find` actually does ("resolves to the latest applicable revision ≤ target, regardless of whether the target itself appears in `revinfo`") and dropping the "between revisions" language. The Description of changes in the PR has the same conflation — gap revisions weren't rejected previously, so that's not the actual fix.

oleksii-donets (commented):
Behaviour-aligning change with a regression worth confirming before merging.

The PR description frames this as fixing a problem where rollback rejected revisions that the snapshot endpoint accepted (gap revisions belonging to other resources). After tracing the code, I don't think gap revisions were actually rejected by the old precheck — `historyService.getRevisionById` checks the `revinfo` table, which receives a row for every audited transaction regardless of which entity changed, so a gap revision created by an unrelated entity's modification already passed the precheck. The actual class of revisions the precheck rejected is revision ids with **no row in `revinfo` at all** (typos, far-future values).

This matters for `DeploymentService.rollback` specifically: with the precheck gone, an out-of-range revision id is leniently resolved by `auditReader.find` to the latest entity snapshot, the rollback proceeds, and `reconcileEnvSecrets` then unconditionally nulls every sensitive env var (it has no "snapshot == current" short-circuit). So a fat-fingered revision id will now silently wipe a deployment's secrets where it used to return 404 with no side effects. The whitelist path is fine — its `current.equals(snapshot)` short-circuit absorbs the case — and the image-definition path is mostly fine modulo a possible spurious `buildStatus = NOT_BUILT` reset.

If the env-secret behaviour is intentional, please call it out in the spec; otherwise a targeted `revision <= maxRevisionId()` guard would preserve the gap-revision intent without the typo footgun. Inline comments have the details.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
Pasichniuk: Good catch on the env-secret regression — the secondary footgun is real. Where the framing diverges: the precheck wasn't only rejecting typos. Hibernate's pooled sequence allocator burns a block of ids on JVM restart, so revinfo has legitimate gaps, and the audit UI's snapshotRevision - 1 rollback routinely targets a gap id. That's the use case this PR is fixing — restoring the precheck would reintroduce that bug. Going with your alternative suggestion: adding a maxRevisionId() helper and guarding revision > maxRevisionId() in all three rollback methods. That keeps gap-id tolerance and rejects typos before they reach reconcileEnvSecrets.

src/main/java/com/epam/aidial/deployment/manager/service/ImageDefinitionService.java
Pasichniuk: Same fix applies — the maxRevisionId() guard short-circuits before updateImageDefinition runs, so no spurious NOT_BUILT reset on out-of-range ids. Keeping the precheck removed because of the sequence-allocator case (see reply on the DeploymentService thread)

src/main/java/com/epam/aidial/deployment/manager/service/GlobalDomainWhitelistService.java
Pasichniuk: Adding the maxRevisionId() guard here too for symmetry with the other two rollback methods, and the spec will explicitly differentiate the three resources' failure modes.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentRollbackFunctionalTest.java
Pasichniuk: Replacing it with two tests per resource: one that synthesises a real allocator-style revinfo gap (insert a revinfo row at currentMax + 50 and roll back to an id in the gap) to lock down the actual behaviour change, and one that asserts out-of-range ids reject with 404 and don't touch the resource.

specs/020-revision-rollback/spec.md
Pasichniuk: Re-anchoring FR-002 on what auditReader.find actually does, and calling out the sequence-allocator motivation explicitly (that's the real reason the precheck had to go — not gap revisions in the cross-entity sense, which already worked). Same edit to the whitelist spec.

ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

PR #339, opened 2026-05-25
fix: Docker Hub and bearer-auth registries on MCP LOCAL builds

### LINKED ISSUE #78: [Images] The Image istallation FAILED. Due to specific authorization on the docker daemon on the pod - Limitation of the downloading images from docker hub (100 times per 6 hours)
### Name and Version

1. Create Image 100+ times per 6 hours
2. Install it

Image
**Installation Failed**

### What do you see instead?

**Installed**

### Additional information
```

Image build started
Image distro analyser is started
[0000] ERROR could not determine source: errors occurred attempting to resolve 'd0nets/simple-mcp:0.0.3':
 - snap: snap file "d0nets/simple-mcp:0.0.3" does not exist
 - docker: docker not available: failed to connect to Docker daemon. Ensure Docker is running and accessible
 - podman: podman not available: no host address
 - containerd: containerd not available: no grpc connection or services is available: unavailable
 - oci-registry: failed to get image descriptor from registry: GET https://index.docker.io/v2/d0nets/simple-mcp/manifests/0.0.3: TOOMANYREQUESTS: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
 - snap: strconv.Atoi: parsing "0.0.3": invalid syntax
 - additionally, the following providers failed with file does not exist: docker-archive, oci-archive, oci-dir, singularity, local-file, local-directory
Image build has failed: Image distro analyzation has failed
```

docker image link is being used

oleksii-donets:
### Applicable issues

- fixes #78

### Description of changes

MCP LOCAL transport builds that pulled a source image from Docker Hub
(or any bearer-auth registry like ACR/GHCR/GAR/ECR) failed with 401
even when the registry was configured under
`app.registry.trusted-private-registries`.

Three independent auth paths needed to align:

- **`RegistryService.dockerConfig()`** — Docker CLI / BuildKit hardcode
  the lookup key `https://index.docker.io/v1/` for the Docker Hub
  Official index. The pull-secret now emits that legacy key for the
  three Hub aliases (`docker.io`, `index.docker.io`,
  `registry-1.docker.io`); every other registry keeps the existing
  `https://<host>/v2` form Skopeo and BuildKit both accept.
- **`DockerRegistryClient.getRegistryClient()`** — jib-core
  canonicalizes Docker Hub references to `registry-1.docker.io`, so a
  user-configured `docker.io` trusted entry never matched. Added an
  alias-aware `isSameRegistry` helper used by both the main-registry
  and trusted-private-registries matchers. Switched the credentialed
  branches from `configureBasicAuth()` to `doPullBearerAuth()` so the
  `WWW-Authenticate` challenge is honoured; basic-only registries
  still fall back to Basic.
- **Pipeline `catch (Exception e)` blocks** (`ImageWrapperBuildPipeline`,
  `ImageBuildFromGitPipeline`, `ImageCopyPipeline`) now `log.error`
  the full stack and fall back to `e.getClass().getSimpleName()` when
  `getMessage()` is null, so the next regression doesn't surface as
  the literal string `Image build has failed: null`.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/docker/DockerRegistryClient.java
```diff
@@ -125,25 +124,27 @@ private RegistryClient getRegistryClient(String imageName) {
...
         if (registryProperties.getAuth() == DockerAuthScheme.BASIC
-                && Objects.equals(registryProperties.getUrl(), imageReference.getRegistry())) {
+                && DockerHubAliases.sameRegistry(registryProperties.getUrl(), imageReference.getRegistry())) {
             var credential = Credential.from(registryProperties.getUser(), registryProperties.getPassword());
             registryClientFactory.setCredential(credential);
             var registryClient = registryClientFactory.newRegistryClient();
-            registryClient.configureBasicAuth();
+            // doPullBearerAuth handles the WWW-Authenticate challenge: bearer (Docker Hub, ACR, GHCR, GAR, ECR, …)
+            // negotiates a token using the credentials; basic-only registries fall back to Basic.
+            registryClient.doPullBearerAuth();
```
Pasichniuk: **Likely regression for basic-only registries.**

In jib-core 0.27.3, `doPullBearerAuth()` does **not** fall back to Basic — and the return value is discarded here. From `RegistryClient.doBearerAuth(...)`:

```java
if (!authenticator.isPresent()) {
    eventHandlers.dispatch(LogEvent.debug("server requires basic auth for " + image));
    return false; // server returned "WWW-Authenticate: Basic ..."
}
```

and `RegistryAuthenticator.fromAuthenticationMethod()`:

```java
if (authenticationMethod.matches("^(?i)(basic).*")) {
    return Optional.empty();
}
```

So when the server advertises `WWW-Authenticate: Basic` (self-hosted Distribution v2 with htpasswd, Harbor in basic mode, Artifactory without bearer, …), `doPullBearerAuth()` returns `false` and the client's `authorization` stays `null`. The subsequent `pullManifest` / `pullBlob` calls go out without an auth header, get 401, and `callRegistryEndpoint` re-throws because `isBearerAuth(null)` is false.

Before this PR, `configureBasicAuth()` preemptively set the Basic header and worked against those registries.

Suggested fix — preserve the Hub/ACR/GHCR fix while keeping basic-only registries working:

```suggestion
            // Try bearer first (Docker Hub, ACR, GHCR, GAR, ECR); fall back to Basic when the
            // registry advertises WWW-Authenticate: Basic (jib's doPullBearerAuth returns false
            // and leaves authorization unset in that case, so we must configure it ourselves).
            if (!registryClient.doPullBearerAuth()) {
                registryClient.configureBasicAuth();
            }
```

The new test should also cover this — mock the auth-discovery response as `WWW-Authenticate: Basic` and assert `configureBasicAuth()` is invoked.

src/main/java/com/epam/aidial/deployment/manager/docker/DockerRegistryClient.java
```diff
@@ -125,25 +124,27 @@ private RegistryClient getRegistryClient(String imageName) {
...
-            if (Objects.equals(trustedRegistry.getRegistry(), imageReference.getRegistry())
+            if (DockerHubAliases.sameRegistry(trustedRegistry.getRegistry(), imageReference.getRegistry())
                     && "BASIC".equals(trustedRegistry.getAuthScheme())
                     && trustedRegistry.getUser() != null
                     && trustedRegistry.getPassword() != null) {
                 var credential = Credential.from(trustedRegistry.getUser(), trustedRegistry.getPassword());
                 registryClientFactory.setCredential(credential);
                 var registryClient = registryClientFactory.newRegistryClient();
-                registryClient.configureBasicAuth();
+                registryClient.doPullBearerAuth();
```
Pasichniuk: Same regression as the main-registry branch above — a basic-only trusted private registry will now be left unauthenticated. Apply the same `if (!registryClient.doPullBearerAuth()) registryClient.configureBasicAuth();` pattern here, and consider extracting it to a tiny helper so both branches stay in sync.

specs/container-registry/spec.md
```diff
@@ -83,5 +83,9 @@ Status: **Implemented**
...
+- `DockerRegistryClient.getRegistryClient()` (jib-core path, used for in-manager-process image inspection like inheriting a source image's entrypoint during wrapper builds):
+  - Treats `docker.io`, `index.docker.io`, and `registry-1.docker.io` as the same registry when matching against `app.registry.url` and `app.registry.trusted-private-registries` entries. jib-core canonicalizes any Docker Hub reference to `registry-1.docker.io`, so a user-configured `docker.io` trusted entry must alias-match — otherwise jib falls through to anonymous bearer auth and fails 401 on private repos.
+  - Once credentials are matched, uses jib's `doPullBearerAuth()` (not `configureBasicAuth()`) for the actual auth handshake. Modern registries (Docker Hub, ACR, GHCR, GAR, ECR, …) return a `WWW-Authenticate: Bearer …` challenge and reject a hardcoded `Authorization: Basic …` header on the registry API; `doPullBearerAuth()` performs the token-endpoint round-trip with the credentials and falls back to Basic when a registry advertises `WWW-Authenticate: Basic`.
```
Pasichniuk: This sentence is inaccurate against the actual jib-core 0.27.3 implementation. `doPullBearerAuth()` returns `false` when the registry advertises `WWW-Authenticate: Basic` and does **not** configure basic auth — see `RegistryClient.doBearerAuth(...)` and `RegistryAuthenticator.fromAuthenticationMethod()` (case-insensitive `^basic.*` match returns `Optional.empty()`). Today the return value is discarded in `DockerRegistryClient`, so basic-only registries are left unauthenticated.

Either (preferred) wire up the explicit Basic fallback in `DockerRegistryClient` so the spec stays accurate, or rewrite this line to reflect what the code actually does (`doPullBearerAuth()` returns `false` for basic-only servers and the caller must handle the fallback).

src/main/java/com/epam/aidial/deployment/manager/service/RegistryService.java
```diff
@@ -80,7 +81,13 @@ public String dockerConfig() {
                     String regProtocol = StringUtils.isNotBlank(trustedRegistry.getProtocol())
                             ? trustedRegistry.getProtocol()
                             : "https";
-                    auths.put(API_URL_TEMPLATE.formatted(regProtocol, trustedRegistry.getRegistry()), authConfig);
+                    String key = authKey(regProtocol, trustedRegistry.getRegistry());
+                    if (auths.containsKey(key)) {
+                        log.warn("Multiple trusted-private-registry entries collapse to the same auth key '{}' "
+                                + "(e.g. Docker Hub aliases docker.io / index.docker.io / registry-1.docker.io). "
+                                + "Later entry overwrites the earlier one — verify only one set of credentials is intended.", key);
```
Pasichniuk, line 89: Minor: the wording assumes the collision is between two trusted entries, but this check also fires when a trusted entry collides with the **main** registry's auth key (e.g. `app.registry.url=docker.io` + a trusted entry for `index.docker.io`, or two equivalent host strings for any registry). In that path the trusted entry silently overrides the main credentials, and the current message misdirects operators toward "two trusted entries."

Suggest either widening the message (e.g. "Multiple registry entries (main or trusted) collapse to the same auth key…") or checking the colliding key against `registryProperties.getUrl()` explicitly to log a more specific message in that case.

src/test/java/com/epam/aidial/deployment/manager/docker/DockerRegistryClientTest.java
```diff
@@ -178,6 +178,50 @@ void getRegistryClient_shouldNotSetCredentialsWhenRegistryDoesNotMatch() throws
...
+            ReflectionTestUtils.invokeMethod(client, "getRegistryClient", "d0nets/private-simple-mcp");
+
+            var credentialCaptor = ArgumentCaptor.forClass(com.google.cloud.tools.jib.api.Credential.class);
+            verify(registryClientFactory).setCredential(credentialCaptor.capture());
+            assertThat(credentialCaptor.getValue().getUsername()).isEqualTo("hubuser");
+            assertThat(credentialCaptor.getValue().getPassword()).isEqualTo("hubtoken");
+
+            // Critical: bearer flow (Docker Hub rejects plain Basic on the registry API).
+            verify(registryClient).doPullBearerAuth();
+        }
```
Pasichniuk, line 275: Worth tightening with `verify(registryClient, never()).configureBasicAuth();` — that makes the regression intent ("Docker Hub rejects preemptive Basic") explicit and lets the suite catch a future revert. Pairs well with adding a second test for the basic-only branch (see the comment on `DockerRegistryClient.java` line 133) — that one would exercise `configureBasicAuth()` being called as the fallback after `doPullBearerAuth()` returns false.

specs/container-registry/spec.md
```diff
@@ -83,5 +83,9 @@ Status: **Implemented**
   - `app.registry.user` / `app.registry.password` — primary registry credentials
   - `app.registry.trusted-private-registries` — JSON array of additional registries with optional BASIC credentials
 - Image copy tool: Skopeo (external process, invoked from build pipeline steps)
+- Pull-secret `config.json` keys (`RegistryService.dockerConfig()`): `https://<host>/v2` for every registry, **except** Docker Hub aliases (`docker.io`, `index.docker.io`, `registry-1.docker.io`) which use the legacy key `https://index.docker.io/v1/`. Skopeo/containers-image normalizes both forms transparently, but `docker/cli` (and therefore BuildKit's authprovider, used by MCP LOCAL transport) hardcodes the legacy key for the Official Docker Hub index — a `/v2`-shaped key for `docker.io` is silently ignored there. If multiple trusted-private-registry entries collapse to the same auth key (e.g. one entry for `docker.io` and another for `index.docker.io`), only the last set of credentials survives; `RegistryService` logs a `WARN` when this happens so operators can spot the misconfiguration in logs.
```
Pasichniuk: Nit: this is one very long single-line bullet that's hard to read in the rendered spec and in plain text. Consider breaking it into sub-bullets matching the structure of the next bullet (the `DockerRegistryClient.getRegistryClient()` one), e.g. one sub-bullet for the key shape, one for the docker/cli rationale, and one for the collision/WARN behavior. Easier to keep in sync with the code, and easier to skim.

Pasichniuk (changes requested):
Solid fix for the Docker Hub / bearer-auth case overall — the alias-aware matching and the `DockerHubAliases` helper are clean. One material concern about the `doPullBearerAuth()` switch (looks like a regression for basic-only registries) plus a few smaller notes inline.

src/main/java/com/epam/aidial/deployment/manager/docker/DockerRegistryClient.java
oleksii-donets: Good catch — fixed in b5a116c3. Extracted a small helper so both this branch and the trusted-registry branch stay in sync:

```java
private static void authenticate(RegistryClient registryClient) throws Exception {
    if (!registryClient.doPullBearerAuth()) {
        registryClient.configureBasicAuth();
    }
}
```

Also added the basic-only fallback test you suggested — `getRegistryClient_shouldFallBackToBasicAuthWhenRegistryAdvertisesBasicChallenge` stubs `doPullBearerAuth()` to return `false` and verifies `configureBasicAuth()` is invoked.

src/main/java/com/epam/aidial/deployment/manager/docker/DockerRegistryClient.java
oleksii-donets: Done in b5a116c3 — same `authenticate()` helper used here (see reply on the main-registry branch above), so the two sites can't drift apart.

specs/container-registry/spec.md
oleksii-donets: Took the preferred option — wired up the explicit Basic fallback in `DockerRegistryClient` (new `authenticate()` helper) and rewrote this sub-bullet in b5a116c3 to match what the code actually does: `doPullBearerAuth()` returns `false` for `WWW-Authenticate: Basic` and `DockerRegistryClient` then calls `configureBasicAuth()` so basic-only registries (self-hosted Distribution v2 + htpasswd, Harbor basic, Artifactory without bearer) still authenticate.

src/main/java/com/epam/aidial/deployment/manager/service/RegistryService.java
oleksii-donets: Fair point — widened the wording in b5a116c3:

```
Multiple registry entries (main or trusted-private) collapse to the same auth key '{}'
(e.g. Docker Hub aliases docker.io / index.docker.io / registry-1.docker.io,
or equivalent host strings for any registry).
Later entry overwrites the earlier one — verify only one set of credentials is intended.
```

Took the simpler of the two options. Happy to switch to the explicit main-vs-trusted check for a more specific diagnostic if you'd prefer.

src/test/java/com/epam/aidial/deployment/manager/docker/DockerRegistryClientTest.java
oleksii-donets: Done in b5a116c3 — added both:

- `verify(registryClient, never()).configureBasicAuth();` on this test (plus `when(registryClient.doPullBearerAuth()).thenReturn(true)` so the new bearer-success path stays exercised and the fallback doesn't kick in).
- New sibling `getRegistryClient_shouldFallBackToBasicAuthWhenRegistryAdvertisesBasicChallenge` that stubs `doPullBearerAuth()→false` and verifies `configureBasicAuth()` is called.

specs/container-registry/spec.md
oleksii-donets: Done in b5a116c3 — split into three sub-bullets mirroring the structure of the `DockerRegistryClient.getRegistryClient()` bullet below:

- Key shape (`/v2` vs. legacy Hub key)
- Why the legacy key matters (`docker/cli` + BuildKit authprovider hardcodes it for the Hub Official index)
- Collision behaviour (Hub aliases + main-vs-trusted, last-wins, `WARN`)

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

PR #346, opened 2026-05-27
feat: cilium policy augmentation for chained predictor + transformer

### LINKED ISSUE #87: [Model Servings] Support text classification models

oleksii-donets:
### Applicable issues

- related to #87

### Description of changes

Implements spec 022 — augments the per-deployment `CiliumNetworkPolicy`
for chained KServe predictor + transformer inference deployments
(introduced by spec 021 / PR #334). Without this, chained deployments
hang in `PENDING` on clusters with Cilium policies enforced because the
transformer → predictor in-cluster hop is blocked at the network layer.

The chained-mode signal flows from `InferenceDeploymentManager`'s
existing deploy-time HuggingFace detection result via a new
`DeployContext` record into the policy creation site — no
persistence, no second HF Hub fetch, no manifest re-parsing on the
deploy / rollingUpdate paths. On the `updateCiliumNetworkPolicy(id)`
path (operator edits `allowedDomains` without changing topology) the
chained flag is sourced from the live KServe `InferenceService`
resource — a Kubernetes API read, not an external upstream fetch. The
endpoint fails fast with `DeploymentException` if the cluster resource
is unexpectedly absent, rather than silently downgrading the policy
(which would recreate the very bug this PR fixes — see
`specs/022-transformer-cilium-policies/spec.md` § Risks). The signal is
task-agnostic: any current or future chained inference task that emits
a `transformer` block triggers the augmentation.

When chained, the `CiliumNetworkPolicy` gains three additions on top
of the baseline:
- A new egress block whose `toEndpoints` list has **six narrowed
  per-app selectors** (no whole-namespace grants): same-`InferenceService`
  pods; `app=istiod` and `app=istio-ingressgateway` in `istio-system`;
  `app=activator`, `app=autoscaler`, `app=controller` in `knative-serving`.
  No `toPorts` — intra-cluster control-plane traffic is unrestricted on
  the port axis. (The original reference YAML used whole-namespace
  selectors; PR-#346 review narrowed to per-app for least-privilege and
  the alignment iteration added `istio-ingressgateway` + `controller`
  back per the reference YAML.)
- Same-`InferenceService` entry appended to the existing ingress
  `fromEndpoints` (alongside `istio-ingressgateway` / `activator` /
  `autoscaler`).
- `8080/TCP` admitted into the ingress port list **via container-port
  resolution** — `InferenceDeploymentManager.getCiliumIngressPorts`
  injects `DEFAULT_KSERVE_SERVICE_PORT = 8080` through the existing
  `ports` channel. No chained-mode literal is injected by
  `CiliumNetworkPolicyCreator`.

When `allowedDomains` is empty (security lockdown) on a chained
deployment, a cluster-DNS-only kube-dns egress block
(`matchPattern: *.svc.cluster.local`) is emitted so the transformer can
still resolve the predictor's cluster-local DNS name. External DNS
resolution remains blocked. When `allowedDomains` is non-empty but
specific (e.g. the default HuggingFace allowlist), the kube-dns block
carries both the per-domain `matchName` entries **and** the
`*.svc.cluster.local` matchPattern, so external HF traffic and
cluster-local predictor resolution both work under Cilium's DNS-proxy
enforcement.

Every non-chained code path (Knative, NIM, image-build jobs, predictor-
only inference) keeps using the existing 5-arg
`CiliumNetworkPolicyCreator.create(...)` overload — verified
byte-identical by all existing tests passing unmodified plus an
explicit recursive-equality canary
(`shouldProduceIdenticalPolicyTo5ArgOverload_whenChainedFalse`).

The `updateCiliumNetworkPolicy(id)` path (called when an operator
edits `allowedDomains`) was extended the same way so the augmentation
is preserved across domain-whitelist edits on chained deployments.
Topology flips on running deployments reach the policy via
`rollingUpdate`'s `afterCommit` block, which now refreshes the CNP
**before** updating the InferenceService to keep the two crash-consistent.

### Runtime network posture side-effect (operators please read before flipping `app.cilium-network-policies-enabled`)

This PR widens when the per-deployment `CiliumNetworkPolicy` is
rewritten, and pairs that with a 404-tolerant create fallback. The
combined effect is intentional but visible to operators of running
deployments:

- **`rollingUpdate` now refreshes the CNP** on `envs` / `resources` /
  `scaling` / `source` / `nodePool` edits. Pre-PR it never touched the
  CNP; only the explicit `allowedDomains`-edit endpoint did.
- **`AbstractDeploymentManager.updateCiliumNetworkPolicy(...)` now
  tolerates a 404 from `.update()`** and falls through to
  `.createCiliumNetworkPolicy(...)` + disposable-resource registration.

Together, this means: a deployment created while
`app.cilium-network-policies-enabled=false` (no CNP, allow-all
networking) will get a real CNP installed mid-flight on its next
routine edit if the flag is later flipped on. The newly-installed CNP
honours the deployment's `allowedDomains` — so existing egress to
domains *not* on the allowlist (likely, since the whitelist was
irrelevant when the deployment was created) will break, and ingress
narrows to the istio-ingressgateway / activator / autoscaler sources.

This is the intended posture per spec 022's Risks section ("closing
the gap") — flag-on means flag-on consistently — but it surfaces on
the *next edit* rather than only on a fresh deploy. Operators flipping
the flag on a running cluster should either (a) trigger a controlled
re-deploy on each existing workload to apply the lockdown
deliberately, or (b) treat the next edit as a maintenance event and
audit `allowedDomains` first. Worth a line in the release notes.

The flag-on-but-Cilium-not-installed misconfiguration is covered
in `specs/022-transformer-cilium-policies/spec.md` § Risks (the
fallback `.create()` 404s and propagates as `DeploymentException`).

### Bundled cleanup

This PR also removes superseded speckit artifacts from spec 021
(`specs/021-inference-task-transformer/{contracts,data-model,plan,quickstart,tasks}.md`,
commit `34403845`). In hindsight this should have been a separate PR;
future doc-cleanup commits will be split out. The deletions are listed
here for transparency:
- `specs/021-inference-task-transformer/contracts/inference-deployment-api.md`
- `specs/021-inference-task-transformer/data-model.md`
- `specs/021-inference-task-transformer/plan.md`
- `specs/021-inference-task-transformer/quickstart.md`
- `specs/021-inference-task-transformer/tasks.md`

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (commented):
## Review — chained predictor + transformer Cilium policy

Well-structured change overall: the `buildCiliumNetworkPolicy` hook keeps the chained/KServe concept out of the abstract base, the signal-threading rework is sound, fail-fast on an absent cluster service is the right call, and the capability specs are updated in lockstep. The 8080 dedup, the deploy/rollingUpdate-read-fresh-spec vs update-path-read-live split, and the `DeploymentService` skip-when-rolling-update guard are all correct and well covered by tests.

One likely-serious correctness gap (inline on `CiliumNetworkPolicyCreator`) plus a medium new failure mode on the rolling-update path and a few low items below. Posting as neutral comments — happy to re-file as request-changes if the High finding is confirmed against a real cluster.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
```diff
@@ -67,10 +80,25 @@ public CiliumNetworkPolicy create(@NotNull String namespace,
...
+        boolean hasAllowedDomains = CollectionUtils.isNotEmpty(allowedDomains);
+        if (hasAllowedDomains) {
+            egressList.add(createExternalEgress(allowedDomains));
+            egressList.add(createKubeDnsEgress(allowedDomains));
+        } else if (chainedTransformer) {
+            // Chained pair uses cluster-local DNS to resolve *-predictor.<ns>.svc.cluster.local.
+            // When external egress is locked down (allowedDomains empty) we still need kube-dns
+            // reachable, but only for cluster-internal names (no external DNS exfil).
+            egressList.add(createClusterDnsOnlyEgress());
         }
```
Pasichniuk, line 99: **🔴 High — cluster-local DNS is missing for the *default* chained case.**

`*.svc.cluster.local` is added **only** in the `else if (chainedTransformer)` branch, i.e. when `allowedDomains` is empty. When `allowedDomains` is non-empty *and not* `["*"]`, `createKubeDnsEgress(allowedDomains)` emits DNS rules that allowlist only those external domains (`toDnsList` → `matchName` per domain). Cilium's DNS proxy then refuses `<svc>-predictor.<ns>.svc.cluster.local`.

The default effective domain set for HF inference is `huggingface.co,transfer.xethub.hf.co,cas-server.xethub.hf.co` (`application.yml:37`, merged in `InferenceDeploymentManager.getEffectiveDeploymentAllowedDomains`) — specific, non-wildcard. So the **common production chained deployment** hits the `hasAllowedDomains` branch and never gets the cluster-local pattern → transformer can't resolve the predictor → deployment hangs in `PENDING`. That's the exact failure SC-001/FR-001 claim to fix, and the spec's own Risk note ("the transformer resolves `*-predictor.<ns>.svc.cluster.local` via kube-dns before reaching the istio mesh") applies here too.

Tests mask it: `CiliumNetworkPolicyCreatorTest` only covers chained + `["*"]` (wildcard `matchPattern: *` happens to cover cluster-local) and chained + `[]`. There's no chained + specific-domains test.

**Suggested fix:** always append `CLUSTER_LOCAL_DNS_PATTERN` to the kube-dns DNS rules when `chainedTransformer` is true, regardless of `allowedDomains`, and add a `chained + List.of("huggingface.co")` test asserting the pattern is present. Worth confirming on a real Cilium cluster before merge.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
```diff
@@ -163,22 +187,81 @@ private List<Dns> toDnsList(List<String> domains) {
...
+
+        ToPorts kubeDnsToPorts = new ToPorts();
+        kubeDnsToPorts.setPorts(List.of(kubeDnsPorts));
+        kubeDnsToPorts.setRules(rules);
+
+        Egress egress = new Egress();
+        egress.setToEndpoints(List.of(kubeDnsToEndpoints));
+        egress.setToPorts(List.of(kubeDnsToPorts));
+        return egress;
+    }
```
Pasichniuk: **🟡 Low (simplification).** `createClusterDnsOnlyEgress()` is a near-copy of `createKubeDnsEgress()` (152–173) — identical except the DNS rule list (`matchPattern: *.svc.cluster.local` vs `matchName` per domain). Parametrizing the DNS rules into one helper would remove the duplication and would naturally enable the High-severity fix above (append the cluster-local pattern to the same kube-dns block when chained).

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
```diff
@@ -222,6 +224,12 @@ public Deployment rollingUpdate(String id) {
                 @Override
                 public void afterCommit() {
                     try {
+                        // Refresh CNP BEFORE updateService: on topology flips a half-applied update
+                        // where the new topology is live but the CNP still blocks the new traffic.
+                        // Over-permissive briefly > under-permissive.
+                        updateCiliumNetworkPolicy(deployment, serviceSpec, id, deployment.getServiceName(),
+                                getEffectiveDeploymentAllowedDomains(deployment),
+                                getCiliumIngressPorts(deployment));
                         updateService(namespace, serviceSpec);
```
Pasichniuk: **🟠 Medium — `rollingUpdate` now issues a CNP `update()` that 404s if the policy is absent.** `K8sClient.updateCiliumNetworkPolicy` uses Fabric8 `.update()` (replace), which throws on a missing resource. Pre-feature, `rollingUpdate` never touched the CNP. New failure mode: a deployment created while `cilium-network-policies-enabled=false` (no CNP) and the flag later turned on → the next rolling update (now triggered by `scaling`/`resources`/`envs`/`source`/`nodePool` — a much wider surface than the old allowedDomains-only path) tries to replace a non-existent CNP, throws in `afterCommit`, and leaves the deployment stuck in `PENDING` with the service never updated. Consider create-or-replace (or tolerate-404 → create) on this refresh path.

**🟡 Low (FR-011).** The CNP is replaced *before* `updateService`. On a `predictor-only → chained` flip where `updateService` then permanently fails, the augmented (over-permissive) CNP is left applied over a still-predictor-only service, with no rollback in the `catch`. The comment frames this as "over-permissive briefly > under-permissive," but "briefly" assumes a retry succeeds — worth acknowledging the no-rollback case, since FR-011 says the policy must not remain augmented without a matching chained InferenceService.

**🟡 Nit (grammar).** The comment reads as missing a verb: "…BEFORE updateService **to avoid** a half-applied update where the new topology is live but the CNP still blocks the new traffic."

CLAUDE.md
```diff
@@ -70,3 +70,10 @@ Where `<slug>` is the directory name under `specs/` that the feature belongs to
...
 3. If the field is `N/A — creates new capability <slug>`, scaffold `specs/<slug>/spec.md` from the house style (use `specs/api-conventions/spec.md` as a model) and add a row to `specs/README.md`.
+
+## Active Technologies
+- Java 21 (constitution §Tech Stack) + Spring Boot 3.5.10, Lombok 8.10 plugin, Fabric8-generated `io.cilium.v2.*` model types (already in the project; data classes only — no K8s CRUD added in service layer), Fabric8 KServe generated types via existing `service/manifest/` code path. (022-transformer-cilium-policies)
+- N/A — feature introduces no schema changes, no persisted columns, no Flyway migration. The chained signal is transient. (022-transformer-cilium-policies)
+
+## Recent Changes
+- 022-transformer-cilium-policies: Added Java 21 (constitution §Tech Stack) + Spring Boot 3.5.10, Lombok 8.10 plugin, Fabric8-generated `io.cilium.v2.*` model types (already in the project; data classes only — no K8s CRUD added in service layer), Fabric8 KServe generated types via existing `service/manifest/` code path.
```
Pasichniuk: Redundant change

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
```diff
@@ -163,22 +187,81 @@ private List<Dns> toDnsList(List<String> domains) {
...
+
+        // Anything in the knative-serving namespace (activator, autoscaler, queue-proxy, webhook).
+        ToEndpoints knativeServing = new ToEndpoints();
+        knativeServing.setMatchLabels(Map.of(KUBE_DNS_NAMESPACE_LABEL_NAME, KNATIVE_NAMESPACE_LABEL_VALUE));
+
+        Egress egress = new Egress();
+        egress.setToEndpoints(List.of(sameInferenceService, istioSystem, knativeServing));
+        // No toPorts — intra-cluster control-plane traffic is unrestricted on the port axis (spec 022 R-004).
+        return egress;
+    }
```
Pasichniuk, line 245: **🟠 Least-privilege — consider narrowing this egress (point A).**

This block grants egress to **all pods** in `istio-system` and `knative-serving` (namespace label only) on **all ports** (no `toPorts`). The endpoint selector matches both the transformer *and* the predictor pod — i.e. the pods running untrusted model code (arbitrary HF models, `image_reference` containers). If a model pod is compromised, this is an all-port reach into two control-plane namespaces — e.g. istiod plaintext xDS (`:15010`), debug/monitoring (`:15014`), knative controller/webhook.

It's egress-only and bounded to these two namespaces + the same deployment (not cluster-wide), and it does mirror the reference YAML — so this isn't blocking. But could you consider scoping it to the components + ports actually needed (e.g. `app=istiod` / `:15012` in istio-system; `app=autoscaler` (+ activator) on their ports in knative-serving) rather than whole-namespace, all-port?

Related question: predictor-only pods don't get this egress today yet also run istio sidecars — do you know why the chained path needs whole-namespace reachability that predictor-only apparently doesn't? That would tell us whether this can be narrowed safely.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
```diff
@@ -207,12 +290,26 @@ private List<Ingress> createIngress(@Nullable Set<Integer> ports) {
...
+                    INGRESS_PORT_8080.equals(p.getPort())
+                            && p.getProtocol() == io.cilium.v2.ciliumnetworkpolicyspec.ingress.toports.Ports.Protocol.TCP);
+            if (!has8080Tcp) {
+                io.cilium.v2.ciliumnetworkpolicyspec.ingress.toports.Ports port8080 =
+                        new io.cilium.v2.ciliumnetworkpolicyspec.ingress.toports.Ports();
+                port8080.setPort(INGRESS_PORT_8080);
+                port8080.setProtocol(io.cilium.v2.ciliumnetworkpolicyspec.ingress.toports.Ports.Protocol.TCP);
+                ingressPorts.add(port8080);
+            }
+        }
```
Pasichniuk: **🟡 Least-privilege — is `8080` in the L4-only rule necessary? (point B).**

`8080` is added to the `toPorts` ingress rule, which carries no `fromEndpoints` — in Cilium that means no L3 source restriction, so `8080` becomes reachable from every in-cluster endpoint (and `world` where routable), not just the chained sources.

But the two real consumers of `8080` — the istio-ingressgateway and the same-`InferenceService` transformer→predictor hop — are already permitted by the *other* ingress rule (`fromEndpoints` with no `toPorts` → those sources on all ports, including the same-InferenceService entry you added). So this `8080` entry looks broader than the requirement: it additionally lets any pod in the cluster hit the model server on `8080`, bypassing the gateway.

Could you consider either scoping it as `fromEndpoints:[same-InferenceService] + toPorts:[8080]`, or dropping it (since Rule A already covers the gateway and same-IS on all ports)? Acknowledging this diverges from the reference YAML — flagging it as a least-privilege call for you to weigh.

Pasichniuk (commented):
Two follow-up security/least-privilege questions for your consideration (defense-in-depth, not blocking — both match the reference YAML as written).

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: Addressed in `fc3c3205`.

`CiliumNetworkPolicyCreator` now appends `matchPattern: *.svc.cluster.local` to the kube-dns DNS rule list whenever `chainedTransformer=true`, regardless of the `allowedDomains` shape (the `["*"]` case still uses the wildcard `matchPattern: *` alone — it already covers cluster-local). The default HF configuration with `huggingface.co,transfer.xethub.hf.co,cas-server.xethub.hf.co` now produces a kube-dns rule list containing those three `matchName` entries **plus** the cluster-local matchPattern.

Regression guard: new test `shouldAppendClusterLocalDnsPattern_whenChainedAndSpecificDomains` in `CiliumNetworkPolicyCreatorTest`. Also added `shouldNotAppendClusterLocalDnsPattern_whenSpecificDomainsButNotChained` so the predictor-only path stays byte-equivalent.

Spec 022 updated: clarification Q3, FR-001, Acceptance Scenario 1, Edge Cases, and Risks all reflect the always-on behavior.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: Addressed in `fc3c3205`. `createKubeDnsEgress(List<String>)` and `createClusterDnsOnlyEgress()` collapsed into a single `createKubeDnsEgress(domains, includeClusterLocal)` driven by `toDnsList(domains, includeClusterLocal)`. The unified helper made the High-severity fix above a one-line append in `toDnsList`.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
oleksii-donets: Addressed in `fc3c3205`.

**404 fallback:** `AbstractDeploymentManager.updateCiliumNetworkPolicy(...)` now wraps `k8sClient.updateCiliumNetworkPolicy(...)` in a try/catch. On `KubernetesClientException` with HTTP 404 it falls through to `k8sClient.createCiliumNetworkPolicy(...)` **and** registers the policy in `disposable_resource` so the undeploy cleanup path picks it up later. Non-404 K8s errors still propagate as `DeploymentException` (covered by the new `updateCiliumNetworkPolicy_shouldPropagate_whenK8sUpdateNon404Error` test). The fix lives at the service layer so both the new rollingUpdate refresh and the pre-existing `updateCiliumNetworkPolicy(id)` endpoint benefit.

Spec 022 Risks section updated accordingly.

**Grammar nit:** fixed — comment now reads `// Refresh CNP BEFORE updateService to avoid a half-applied update where ...`.

**FR-011 sub-point (no rollback if `updateService` fails after CNP refresh):** acknowledged but not changed in this PR. It's the same `afterCommit`-no-rollback trade-off used across `deploy()` / `undeploy()` — leaving the augmented CNP briefly out-of-sync until the next retry is consistent with how the rest of the deployment lifecycle handles transactional/cluster-write divergence. Worth tracking as a separate cross-cutting reliability improvement rather than scoping it into spec 022.

CLAUDE.md
oleksii-donets: Removed in `fc3c3205`. (Note: speckit's `/speckit.plan` re-adds this block on the next plan run — that's a workflow-level concern, not something I want to fight in this PR.)

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: Narrowed in `fc3c3205`.

`createChainedIntraClusterEgress` now uses three named-app selectors instead of two whole-namespace selectors:
- `app=istiod` in `istio-system` (xDS sidecar config sync — `istio-ingressgateway` is an ingress source, not an egress target, so it doesn't belong here)
- `app=activator` in `knative-serving` (cold-start data path)
- `app=autoscaler` in `knative-serving` (metrics push)

Same-`InferenceService` toEndpoint unchanged. `toPorts` still unset (spec 022 FR-002 — no port restriction on the chained intra-cluster block).

> predictor-only pods don't get this egress today yet also run istio sidecars — do you know why the chained path needs whole-namespace reachability that predictor-only apparently doesn't?

Right: predictor-only pods don't initiate cluster-network egress to istio-system/knative-serving because their sidecar's xDS connection and the queue-proxy localhost intercept handle in-pod mediation without leaving the pod. The chained pair needs additional egress because the transformer→predictor inter-component routing goes through queue-proxy/activator, which DOES need cross-pod egress to the knative components. `istiod` is needed for sidecar config sync once per pod (this works in predictor-only because the same egress is implicit in some cluster setups, but spelling it out for the chained case is the safer default).

**Risk noted in spec 022 Risks:** if a Knative or Istio sub-component beyond `istiod`/`activator`/`autoscaler` becomes a required egress target (e.g. `net-istio-webhook`, a custom validator), the chained pair will fail closed and the selectors will need to be widened. Worth re-validating against your actual cluster topology in the manual verification pass.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: Partially addressed in `fc3c3205`.

**Minimal fix landed:** the chained branch's explicit 8080 literal + dedup block in `createIngress` is gone. 8080 already arrives via the caller's container-port resolution — `InferenceDeploymentManager.getCiliumIngressPorts(...)` adds `DEFAULT_KSERVE_SERVICE_PORT=8080`, so the net policy is unchanged but the redundant injection is removed. Spec 022 FR-001 wording updated to reflect that 8080 is sourced from container-port resolution, not a chained-mode literal.

**Deeper concern deferred:** moving 8080 out of the L4-only `toPorts` rule (no `fromEndpoints`) into a `fromEndpoints=same-IS + toPorts=[8080]` rule is a pre-existing baseline behavior shared by **every** inference deployment (predictor-only included — `defaultContainerPort=8080` lands there too). Tightening this in spec 022's PR would change predictor-only policy shape and break the SC-002 byte-equivalence canary. It belongs in a separate PR with its own predictor-only impact analysis — flagging as a tech-debt follow-up rather than rolling it in here.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
Pasichniuk: Have you validated it locally?

CLAUDE.md
Pasichniuk: Still there

ai-dial-actions:
<h1>Dependency Review</h1>
None

CLAUDE.md
oleksii-donets: removed

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: It was verified on the dev env.  All the required egres endpoints were clarified in the generated CNP

```   
- matchLabels:
    serving.kserve.io/inferenceservice: "dm-deberta"
- matchLabels:
    k8s:io.kubernetes.pod.namespace: "istio-system"
    app: istiod
- matchLabels:
    k8s:io.kubernetes.pod.namespace: "istio-system"
    app: istio-ingressgateway
- matchLabels:
    k8s:io.kubernetes.pod.namespace: "knative-serving"
    app: "activator"
- matchLabels:
    k8s:io.kubernetes.pod.namespace: "knative-serving"
    app: "autoscaler"
- matchLabels:
    app: "controller"
    k8s:io.kubernetes.pod.namespace: "knative-serving" 
```

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
```diff
@@ -222,6 +224,12 @@ public Deployment rollingUpdate(String id) {
                 @Override
                 public void afterCommit() {
                     try {
+                        // Refresh CNP BEFORE updateService to avoid a half-applied update where
+                        // the new topology is live but the CNP still blocks the new traffic.
+                        // Over-permissive briefly > under-permissive.
+                        updateCiliumNetworkPolicy(deployment, serviceSpec, id, deployment.getServiceName(),
+                                getEffectiveDeploymentAllowedDomains(deployment),
+                                getCiliumIngressPorts(deployment));
```
Pasichniuk: **Behavioral side-effect — please record the decision explicitly in the PR description.** The logic itself is acceptable, but this line plus the 404→create fallback below changes the runtime network posture of *running* workloads, for **all** deployment types, as a side-effect of unrelated edits:

- `rollingUpdate` now rewrites the CNP on every applicable edit (envs, resources, scaling, source, nodePool) — pre-PR it never touched the CNP.
- Combined with the create-on-404 fallback, a deployment created while `app.cilium-network-policies-enabled=false` (no CNP, unrestricted networking) silently gets a lockdown policy installed mid-flight on a routine env edit. If it talks to domains not in its `allowedDomains` (likely, since domains were irrelevant when it was created), its egress breaks; ingress narrows to the istio-gateway/activator/autoscaler sources.

Spec 022's Risks section documents this as intentionally "closing the gap", which is defensible — but it deserves a deliberate, visible decision: please add a paragraph to the PR description stating that this is intended, covering the flag-flip scenario above, and flag it for the release notes / upgrade plan so operators flipping the flag on know existing RUNNING deployments will be locked down on their next edit rather than on redeploy.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
```diff
@@ -787,33 +806,51 @@ private void createCiliumNetworkPolicy(String groupId, List<String> allowedDomai
...
+            // rather than 404-ing. Register the disposable resource so undeploy cleanup picks it
+            // up later.
+            log.info("updateCiliumNetworkPolicy. CNP '{}' not found; creating it instead", cnpName);
+            disposableResourceManager.saveK8sResources(
+                    List.of(ciliumNetworkPolicy), K8sResourceKind.CILIUM_NETWORK_POLICY, groupId, namespace);
+            k8sClient.createCiliumNetworkPolicy(namespace, ciliumNetworkPolicy);
+            disposableResourceManager.changeResourceLifecycleByGroupId(
+                    groupId,
+                    new K8sResourceReference(this.namespace, K8sResourceKind.CILIUM_NETWORK_POLICY, cnpName),
+                    ResourceLifecycleState.STABLE);
```
Pasichniuk, line 860: **404 here is ambiguous: "CNP object missing" vs "CiliumNetworkPolicy CRD not installed".** Fabric8 returns 404 in both cases. On a cluster with the flag on but no Cilium CRD, `update()` 404s → this fallback `create()` also 404s → every rolling update now hard-fails (pre-PR, rolling updates never touched the CNP, so such clusters worked). Note that `deleteCiliumNetworkPolicy` a few lines below explicitly tolerates exactly this case (*"…or its CRD not found in namespace (404). Treating as deleted"*). Consider giving the create the same tolerance (or distinguishing the two 404s), or at least acknowledging the failure mode in spec 022's Risks.

Minor, related: `saveK8sResources` is `REQUIRES_NEW`, so the disposable row commits before `createCiliumNetworkPolicy`. If the create then fails (CRD missing, 409 race, RBAC), a TEMPORARY row for a nonexistent CNP is durably committed — and if a disposable row already existed (CNP hand-deleted from the cluster), this inserts a duplicate. Both appear tolerated by cleanup (delete tolerates 404), and the deploy path has the same shape, so consistency-level only.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
```diff
@@ -67,10 +84,27 @@ public CiliumNetworkPolicy create(@NotNull String namespace,
...
+        if (hasAllowedDomains) {
+            egressList.add(createExternalEgress(allowedDomains));
+            // Chained pairs need to resolve *-predictor.<ns>.svc.cluster.local via kube-dns even
+            // when allowedDomains is a specific list — Cilium's DNS proxy would otherwise reject
+            // cluster-local names not in the matchName set.
+            egressList.add(createKubeDnsEgress(allowedDomains, chainedTransformer));
+        } else if (chainedTransformer) {
+            // External egress locked down; emit a kube-dns block restricted to cluster-local
+            // names only so the predictor's cluster DNS name resolves without external DNS exfil.
+            egressList.add(createKubeDnsEgress(List.of(), true));
```
Pasichniuk, line 98: **The "lockdown stays locked down" framing is inverted for this branch.** Pre-PR, empty `allowedDomains` meant *no egress section at all* — and Cilium only enforces default-deny for directions a policy covers, so egress was effectively **allow-all**, not locked down. By emitting kube-dns + intra-cluster egress rules here, a chained deployment with empty effective domains flips into genuine egress default-deny: among other things the predictor can no longer pull model weights from HF at startup, so the deployment likely never starts (rather than "locked down but working", as spec 022's edge-case section implies).

Reachable only when `defaultAllowedDomains` is configured empty (HF sources merge the defaults in via `InferenceDeploymentManager.getEffectiveDeploymentAllowedDomains`), so narrow — and arguably the right semantics. But the spec/docs should state the actual outcome: *chained + empty domains ⇒ deployment cannot start because model download is blocked*, not just "external DNS resolution remains blocked".

specs/kubernetes-manifests/spec.md
```diff
@@ -84,11 +84,12 @@ For chained deployments:
...
+- The per-deployment `CiliumNetworkPolicy` carries three additions over the predictor-only baseline: an intra-cluster egress block to same-`InferenceService` pods plus `istio-system` and `knative-serving` namespaces (no port constraint); a same-`InferenceService` entry appended to the existing ingress `fromEndpoints`; and `8080/TCP` deduplicated into the ingress `toPorts`. Non-chained deployments produce byte-identical policies to the pre-feature shape. *(Implemented via 022-transformer-cilium-policies — see that spec for FR-level detail. The chained-mode signal is read by `InferenceDeploymentManager`'s override of the `buildCiliumNetworkPolicy(...)` hook on `AbstractDeploymentManager`; the abstract base carries no inference-specific concept.)*
```
Pasichniuk: **Spec text drift vs the shipped implementation** (spec 022 FR-001 and `specs/inference-deployments/spec.md` have it right; this line is stale on two points):

- *"…plus `istio-system` and `knative-serving` namespaces"* — the implementation ships six **narrowed per-app selectors** (istiod, istio-ingressgateway, activator, autoscaler, controller + same-`InferenceService`), not whole-namespace selectors.
- *"`8080/TCP` deduplicated into the ingress `toPorts`"* — the dedup-literal approach was removed; 8080 arrives via the caller's `getCiliumIngressPorts` (`DEFAULT_KSERVE_SERVICE_PORT`), with no chained-mode injection.

(The PR description's first bullet has the same stale "namespaces" wording.)

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
```diff
@@ -222,6 +224,12 @@ public Deployment rollingUpdate(String id) {
...
                     try {
+                        // Refresh CNP BEFORE updateService to avoid a half-applied update where
+                        // the new topology is live but the CNP still blocks the new traffic.
+                        // Over-permissive briefly > under-permissive.
+                        updateCiliumNetworkPolicy(deployment, serviceSpec, id, deployment.getServiceName(),
+                                getEffectiveDeploymentAllowedDomains(deployment),
+                                getCiliumIngressPorts(deployment));
                         updateService(namespace, serviceSpec);
                     } catch (Exception e) {
                         var errorMessage = "Rolling update failed for deployment '%s'".formatted(id);
```
Pasichniuk: Nit: if only the CNP refresh fails here, the service update is skipped, but the operator sees a generic "Rolling update failed". The DB keeps the edit, the cluster gets neither change, and reconcile later flips status back to RUNNING against the old spec — so the edit silently stays unapplied after the 5xx. Consider naming the failed step (CNP refresh vs service update) in the error message to aid debugging.

Pasichniuk (changes requested):
Reviewed the full policy-generation path (deploy / rollingUpdate / `updateCiliumNetworkPolicy(id)`), the creator changes, and the test matrix. The PR does fix the problem it targets: all three network-layer blockers for the chained pair (intra-cluster egress, cluster-local DNS under a specific `allowedDomains` list, same-`InferenceService` ingress + 8080) are addressed, the chained signal threading matches spec 022 FR-005 on every path, and the non-chained baseline is protected by the 5-arg/6-arg recursive-equality canary. Design (subclass `buildCiliumNetworkPolicy` hook) is clean, CI is green.

Two non-inline points:

1. **Issue linkage**: `fixes #87` will auto-close *"[Model Servings] Support text classification models"* — the umbrella feature request whose acceptance criteria include a DIAL KServe adapter discoverable in the Create Entity/Model UI workflow. This PR is the network-policy enabler for one piece of that, and the UI/adapter criteria are out of this repo's scope. Suggest changing to `Related to #87` (or linking a dedicated bug issue) so the merge doesn't close the umbrella issue prematurely. (The PR body / spec 022 Risks also describe #87 as "the chained pair hangs in PENDING bug", which misstates what #87 actually says.)

2. **Description accuracy (nit)**: "verified byte-identical by all existing tests passing **unmodified**" — several existing tests were modified (mock stubs widened to the 6-arg overload, functional-test mocks split, `setServiceName(...)` additions). The byte-equivalence canary is the stronger evidence anyway; suggest rewording.

Finally: unit tests prove the policy *shape*; only the spec's Independent Test on a real Cilium cluster proves the traffic actually flows (Cilium DNS-proxy behavior under search-domain expansion is environment-sensitive). Please confirm that end-to-end verification was run before release.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
oleksii-donets: Addressed in `85aabe11` (PR-description-only; no code change).

The PR description now has a dedicated section **"Runtime network posture side-effect (operators please read before flipping `app.cilium-network-policies-enabled`)"** covering exactly the two changes you flagged:

- `rollingUpdate` rewrites the CNP on `envs` / `resources` / `scaling` / `source` / `nodePool` edits — pre-PR, only the `allowedDomains` endpoint touched the CNP.
- Combined with the `.update()`-404→`.create()` fallback in `AbstractDeploymentManager.updateCiliumNetworkPolicy(...)`, a deployment created while the flag was `false` gets a real CNP installed on its next routine edit if the flag is later flipped on.

The new section also calls out the operator-facing consequence (existing egress to domains *not* in `allowedDomains` breaks; ingress narrows), and flags the need for a release-notes line so operators flipping the flag on a running cluster either trigger a controlled re-deploy or audit `allowedDomains` first.

Spec 022's Risks bullet ("CNP-absent on the update path…") already documents the technical change; the PR description now makes the operator-visible decision explicit as you requested.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
oleksii-donets: Addressed in `85aabe11` (spec only — the failure mode is now documented in spec 022 § Risks; code unchanged).

You're right: the fallback `.create()` implicitly assumes the 404 means "object missing" — if the CRD itself is missing it 404s again and propagates as `DeploymentException`. The new Risks bullet ("404 ambiguity on the create-fallback path (CRD missing vs. object missing)") spells this out: pre-PR-#346 this never surfaced on the rollingUpdate path; post-PR it surfaces on every applicable edit; the delete path tolerates this at the `K8sClient` layer (`K8sClient.deleteCiliumNetworkPolicy` line 240–242 "or its CRD not found in namespace (404). Treating as deleted"). Mitigation: install Cilium before flipping the flag. Distinguishing the two 404 shapes at the `K8sClient` layer and mirroring the delete-path tolerance is tracked as a follow-up rather than scoped into spec 022.

**Re: the `REQUIRES_NEW` ordering sub-note** — agreed it's consistency-level only; the pre-existing deploy path has the same shape (`saveK8sResources` commits before `createCiliumNetworkPolicy`). Cleanup tolerates 404 on delete, so a stale TEMPORARY row for a non-existent CNP is harmless. Not touching this in spec 022's PR.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: Addressed in `85aabe11` (spec only — wording sharpened; behaviour unchanged).

You're right: pre-feature, an empty `allowedDomains` produced no `egress` section at all, which under Cilium's "only-enforce-what-the-policy-covers" semantics meant effectively allow-all egress. Post-feature, the chained branch emits kube-dns + intra-cluster egress rules unconditionally, which flips the chained pair into genuine egress default-deny — and the practical consequence is the predictor cannot reach HuggingFace to download model weights, so the deployment cannot start.

Spec 022 § Edge Cases bullet for `allowedDomains empty on a chained deployment` now states this explicitly:

> *Practical consequence: this is a behaviour change from the pre-feature shape — pre-feature, an empty `allowedDomains` produced no `egress` section at all, which under Cilium meant **allow-all** egress; post-feature, the chained pair flips into genuine default-deny on the egress axis, so the predictor cannot reach HuggingFace to download model weights and the deployment will likely never reach `RUNNING`. Reachable only when `app.default-allowed-domains` is configured empty (the HF source's default-merge in `InferenceDeploymentManager.getEffectiveDeploymentAllowedDomains` normally prevents this), but operators who configure the default empty for security lockdown need to know chained deployments will not start without an explicit allowlist that admits the model source.*

So the spec now matches reality: "locked down" no longer implies "locked down but working".

specs/kubernetes-manifests/spec.md
oleksii-donets: Addressed in `85aabe11`.

Both stale points fixed in `specs/kubernetes-manifests/spec.md`:

1. The bullet under "Chained predictor + transformer for text-classification inference" — the "namespaces" phrasing is replaced with the six narrowed per-app selectors that actually ship (same-`InferenceService`, `app=istiod` + `app=istio-ingressgateway` in `istio-system`, `app=activator` + `app=autoscaler` + `app=controller` in `knative-serving`).
2. The `8080/TCP` phrasing is replaced — `8080/TCP` is now described as admitted "via container-port resolution (`InferenceDeploymentManager.getCiliumIngressPorts` injects `DEFAULT_KSERVE_SERVICE_PORT = 8080`), not via a chained-mode dedup literal".

The Scenario block ("Chained-mode Cilium policy augmentation") had the same drift on the same two points — also updated.

**PR description** — same two stale points were also there. New first bullet enumerates the six per-app selectors; new third bullet sources 8080 from container-port resolution rather than a chained-mode literal.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
oleksii-donets: Addressed in `85aabe11`.

The single `try/catch` inside `rollingUpdate.afterCommit` is split into two named-step blocks:

```java
// Refresh CNP BEFORE updateService to avoid a half-applied update where
// the new topology is live but the CNP still blocks the new traffic.
// Over-permissive briefly > under-permissive.
try {
    updateCiliumNetworkPolicy(deployment, serviceSpec, id, ...);
} catch (Exception e) {
    var errorMessage = "Rolling update failed during Cilium policy refresh for deployment '%s'".formatted(id);
    log.warn(errorMessage, e);
    throw new DeploymentException(errorMessage, e);
}
try {
    updateService(namespace, serviceSpec);
} catch (Exception e) {
    var errorMessage = "Rolling update failed during service update for deployment '%s'".formatted(id);
    log.warn(errorMessage, e);
    throw new DeploymentException(errorMessage, e);
}
```

So the 5xx now names the failing step ("Cilium policy refresh" vs "service update"). `InferenceDeploymentManagerTest.rollingUpdate_shouldHandleExceptionDuringUpdate` (which mocks `updateService` to throw) was updated to pin the new message text.

**Re: the wider silent-staleness concern** (DB keeps the edit, reconcile flips status back to RUNNING against the old spec after a 5xx) — agreed it's real but it's a pre-existing reconcile-loop pattern shared across every deployment kind, not introduced by this PR. Worth tracking as a separate cross-cutting reliability ticket rather than scoping it into spec 022.

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
Pasichniuk: One small correction to the new edge-case text added in `85aabe11`: it cites `app.default-allowed-domains`, but the actual property is `app.huggingface.default-allowed-domains` (env `HUGGINGFACE_DEFAULT_ALLOWED_DOMAINS`, `application.yml` line 37). Worth fixing so operators grepping for the key find the right one. Otherwise the rewritten framing matches the implementation — resolved from my side once the property name is corrected.

Pasichniuk:
One leftover description nit from the review summary: the "Description of changes" section still says the non-chained paths are *"verified byte-identical by all existing tests passing **unmodified**"* — but several existing tests were modified in this PR (mock stubs widened to the 6-arg `CiliumNetworkPolicyCreator.create` overload, the `DeploymentFunctionalTest` mock split + `setServiceName(...)` additions, and the error-message assertion re-pinned in `85aabe11`). The recursive-equality canary (`shouldProduceIdenticalPolicyTo5ArgOverload_whenChainedFalse`) is the actual byte-equivalence evidence — suggest rewording the sentence to lean on that instead.

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
oleksii-donets: Fixed in `341e5e81`. The Edge Cases bullet for `allowedDomains` empty on a chained deployment now cites `app.huggingface.default-allowed-domains` (env `HUGGINGFACE_DEFAULT_ALLOWED_DOMAINS`, `application.yml` line 37) — matches the actual Spring property path under `app.huggingface.*`. Repo-wide grep for the old `app.default-allowed-domains` string now returns zero hits.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

PR #351, opened 2026-06-03
feat: entity rollback resurrection

### LINKED ISSUE #350: Re-create image definition / deployment on rollback
### Name and Version

ai-dial-admin-deployment-manager-backend 0.17.0

Currently, on image definition / deployment rollback, if provided id doesn't have a corresponding entity in DB - error is thrown. Problem: entity with that id might have existed before current revision, and user wants to revert it (which means re-creating it).

**Implement rolling back a deleted resource**: If the addressed `{id}` has been deleted but a snapshot exists at-or-before the target revision, rollback should **re-create** the resource instead of returning 404. For **deployments** the original id should be preserved (client-assigned ids); for **image definitions** a NEW server-generated id should be assigned (the UUID generator overwrites any supplied id) and the snapshot's `(name, version)` must be free (a collision rejects with HTTP 409). The **whitelist** is a singleton with no per-id deletion, so this case does not apply to it.

_No response_

Pasichniuk:
### Applicable issues

- fixes #350 

### Description of changes

  - Rolling back a deleted resource now re-creates it instead of returning 404. When a deployment or image definition no longer exists under the addressed {id} but has an audit snapshot at-or-before the target revision, rollback resurrects it from that snapshot in a single call — turning the rollback endpoint into a "revert a deletion" operation. (A truly never-existed id still 404s.)
  - Identity differs by resource type: resurrected deployments keep their original (client-assigned) id, while image definitions come back under a new server-generated UUID — so clients must read the returned DTO for the new id. Image-def re-creation is still subject to the (name, version) uniqueness check (HTTP 409 on collision).
  - The deleted generation's leftovers are drained first. A new ComponentCleanupService.finalizePendingCleanup synchronously cleans lingering K8s/disposable resources and drops any pending component-removal marker before re-creating, so the resurrected resource doesn't collide with resources awaiting deletion and isn't later re-deleted by the scheduled cleaner. Resurrected resources come back inert (NOT_DEPLOYED / NOT_BUILT), with sensitive env values provisioned empty (audit never stored them) for the operator to re-supply.
  - Specs, OpenAPI docs, and functional tests updated across 020-revision-rollback, deployments, and image-definitions, plus controller response codes (added 409) and new tests covering resurrection, removal-marker drain, and name+version collision.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/dao/repository/ComponentRemovalRepository.java
```diff
@@ -35,4 +35,12 @@ public void delete(String id, ComponentType type) {
         jpaRepository.deleteById(new ComponentId(id, type));
     }
 
+    /** Idempotent counterpart to {@link #delete}, which throws when the row is absent. */
```
oleksii-donets: `deleteIfPresent` is redundant, and the javadoc on it is factually wrong.

  In Spring Data Commons 3.x (you're on Spring Boot 3.5.10), the `CrudRepository.deleteById` contract is:

  > If the entity is not found in the persistence store it is silently ignored.

  And `SimpleJpaRepository.deleteById` implements that exactly:

  ```java
  public void deleteById(ID id) {
      Assert.notNull(id, ID_MUST_NOT_BE_NULL);
      findById(id).ifPresent(this::delete);
  }
 ```

  So delete(id, type) does not throw when the row is absent — it's already idempotent.

  `deleteIfPresent` then makes things slightly worse: it does an existsById (one SELECT) followed by deleteById (which internally does another findById SELECT, then DELETE) — two reads where one would do. Functionally equivalent to just calling deleteById.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -190,6 +197,42 @@ public Deployment rollback(String id, Integer revision) {
...
+     * never stores sensitive env values — sensitive envs are provisioned empty and must be re-supplied
+     * via {@code updateDeployment} before deploying, symmetric with {@link #reconcileEnvSecrets}.
+     */
+    private Deployment resurrect(String id, Integer revision) {
+        var snapshot = getDeploymentSnapshot(id, revision);
+        resolveImageDefinitionReference(snapshot);
+        componentCleanupService.finalizePendingCleanup(id, ComponentType.DEPLOYMENT);
+
+        var request = deploymentMapper.toCreateDeployment(snapshot);
+        // toCreateDeployment drops nodePoolId; restore it verbatim (false = no create-time cascade).
```
oleksii-donets: This comment is wrong on two counts, and the line below it is redundant.

1. `toCreateDeployment` does **not** drop `nodePoolId`. The mapper declaration at `DeploymentMapper.java:62` has no `ignore` for the field, and both `Deployment` and `CreateDeployment` expose `nodePoolId`, so MapStruct copies it (confirmed in the generated `DeploymentMapperImpl` — every subclass mapping assigns `createXxxDeployment.nodePoolId(xxxDeployment.getNodePoolId())`). The explicit `request.setNodePoolId(snapshot.getNodePoolId())` on the next line is therefore a no-op.
2. The "(false = no create-time cascade)" parenthetical describes the `false` on line 219 (`createDeployment(request, false)`), not the `setNodePoolId` line it sits above. And even moved to line 219 it would still be redundant: the parameter is named `applyNodePoolDefaultIfEmpty`, which is self-documenting at the call site. "Cascade" is internal vocabulary of `applyCreateTimeNodePoolCascade` and doesn't need to leak here.

Suggested cleanup — delete both the comment and the redundant assignment:

```suggestion
    var request = deploymentMapper.toCreateDeployment(snapshot);
    hydrateEnvValuesFromSnapshot(request, snapshot.getEnvs());
    return createDeployment(request, false);
```

The method's Javadoc already covers the *why* of resurrection; `applyNodePoolDefaultIfEmpty=false` at the call site covers the *why* of skipping the default.

src/main/java/com/epam/aidial/deployment/manager/service/ImageDefinitionService.java
```diff
@@ -173,6 +178,31 @@ public ImageDefinition rollback(UUID id, Integer revision) {
...
+    private ImageDefinition resurrect(UUID id, Integer revision) {
+        var snapshot = getImageDefinitionSnapshot(id, revision);
+        var snapshotType = ImageType.of(snapshot);
+        imageDefinitionRepository.getImageDefinitionByTypeAndNameAndVersion(snapshotType, snapshot.getName(), snapshot.getVersion())
+                .ifPresent(existing -> {
+                    throw new EntityAlreadyExistsException(
+                            "Image definition already exists: type=%s, name=%s, version=%s"
+                                    .formatted(snapshotType, snapshot.getName(), snapshot.getVersion()));
+                });
+        componentCleanupService.finalizePendingCleanup(String.valueOf(id), ComponentType.IMAGE_DEFINITION);
```
oleksii-donets: This call is a no-op for the resurrected entity and the Javadoc above overstates what it achieves. Walk through the three things `finalizePendingCleanup(A, IMAGE_DEFINITION)` does, all keyed by the old id `A`:

1. `markResourcesForCleanupByGroupId(A)` / `cleanAllCleanableByGroupId(A)` — clean A's leftover disposable resources. Disposable resources are group-id'd by entity id, and the resurrected entity gets a **fresh UUID `B`** (see `setId(null)` below). Nothing here touches `B`.
2. `componentRemovalRepository.deleteIfPresent(A, ...)` — drops `A`'s pending-removal marker. If we skipped this, the scheduled cleaner would later run `ImageDefinitionCleanupStrategy.delete(A)` against id `A`. `B` is at a different pk, so the strategy can't reach `B` regardless.

Also note: `createImageDefinition` has no `checkNoResourcesAreAssociatedWithId`-style guard (compare `DeploymentService.createDeployment:140`), so even pre-existing leftovers for `A` wouldn't block the insert of `B`.

Contrast with `DeploymentService.resurrect`, where the drain is genuinely load-bearing on two fronts: deployment id is preserved across resurrect, so without it the scheduled cleaner would re-delete the resurrected row, and the no-leftover-resources guard would refuse the create. Neither failure mode exists here because the id changes.

Net effect of this line: it only changes *when* `A`'s dead leftovers get cleaned (now, synchronously) vs. *whenever the scheduler next runs*. It doesn't protect `B`.

Suggest deleting the call and rewording the Javadoc paragraph that claims it prevents cleaner re-deletion. If we ever switch image-def to id-preserving resurrect for parity with deployment, the drain becomes load-bearing again — worth a one-line breadcrumb in `createImageDefinition` then, but not now.

specs/020-revision-rollback/spec.md
```diff
@@ -93,7 +93,7 @@ A platform administrator notices that a recent change to the global image-build
...
...to a `(name, version)` uniqueness check — see the respective capability specs), (b) the supplied revision identifier is missing, non-positive, or greater than the highest assigned revision id, or (c) the resource has no recorded state at-or-before the supplied revision. For an in-range revision (per (b)) that resolves to a recorded snapshot, the system MUST resolve the target through Envers's point-in-time semantics — to the resource's state as of the latest applicable revision ≤ the target. This means revision ids that fall in `revinfo` gaps are honoured. The motivating case is Hibernate's pooled sequence allocator, which burns a block of revision ids on every JVM restart and so leaves `revinfo` with gap ids that no transaction ever wrote a row for — the audit UI's "roll back to one before this snapshot" pattern (`snapshotRevisionId - 1`) routinely targets such ids. The upper-bound check on (b) exists specifically to keep typo'd or far-future revision ids from silently resolving to the resource's current state and re-applying it (which on the deployment path would wipe live sensitive env values, and on the image-definition path would spuriously reset `buildStatus` to `NOT_BUILT`).
```
oleksii-donets: FR-002 (a) and (c) are now redundant.

After the change:
- (a) fires on "doesn't currently exist **and** has no recorded state at-or-before the revision"
- (c) fires on "has no recorded state at-or-before the revision"

(a) is a strict subset of (c) — any input that satisfies (a) also satisfies (c). The "currently exists" half of (a) no longer carries weight, because a currently-existing resource with no snapshot still 404s under (c), and a currently-missing resource *with* a snapshot is now re-created (handled in the prose, not as a 404 condition).

Suggest collapsing into one clause:

```suggestion
- **FR-002**: The system MUST reject the rollback with HTTP 404 when (a) the supplied revision identifier is missing, non-positive, or greater than the highest assigned revision id, or (b) the resource has no recorded state at-or-before the supplied revision. When the resource does not currently exist under the supplied identifier but a snapshot does exist at-or-before the revision, the system re-creates the resource instead of returning 404 — deployments under their original id; image definitions under a new server-generated id, subject to a `(name, version)` uniqueness check (see the respective capability specs). For an in-range revision (per (a)) that resolves to a recorded snapshot, the system MUST resolve the target through Envers's point-in-time semantics — to the resource's state as of the latest applicable revision ≤ the target. This means revision ids that fall in `revinfo` gaps are honoured. The motivating case is Hibernate's pooled sequence allocator, which burns a block of revision ids on every JVM restart and so leaves `revinfo` with gap ids that no transaction ever wrote a row for — the audit UI's "roll back to one before this snapshot" pattern (`snapshotRevisionId - 1`) routinely targets such ids. The upper-bound check on (a) exists specifically to keep typo'd or far-future revision ids from silently resolving to the resource's current state and re-applying it (which on the deployment path would wipe live sensitive env values, and on the image-definition path would spuriously reset `buildStatus` to `NOT_BUILT`).
```

src/main/java/com/epam/aidial/deployment/manager/dao/repository/ComponentRemovalRepository.java
Pasichniuk: Accepted — verified against the resolved spring-data-jpa 3.5.11 bytecode: `deleteById` is `findById(id).ifPresent(this::delete)`, so `delete` was already idempotent. Removed `deleteIfPresent`, switched `finalizePendingCleanup` to `delete`, and corrected the javadoc in e6c42eb.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
Pasichniuk: Accepted — confirmed in the generated `DeploymentMapperImpl` that every `Deployment → CreateXxxDeployment` mapping copies `nodePoolId`, so the reassignment was a no-op and the comment wrong on both counts. Applied the suggested cleanup verbatim in e6c42eb.

src/main/java/com/epam/aidial/deployment/manager/service/ImageDefinitionService.java
Pasichniuk: Accepted — your walkthrough checks out: all image-def leftovers (jobs `base-/wrapper-/copy-<id>`, configmaps, pushed registry images) are keyed by the old id, the resurrected definition gets a fresh UUID, `createImageDefinition` has no leftover-resource guard, and the scheduled cleaner handles a missing old row gracefully (`getImageDefinitionForUpdateById` returns Optional, `deleteById` no-ops). Removed the call, rewrote the javadoc (with the id-preserving breadcrumb), and repurposed the drain test to assert the resurrected definition survives a `ScheduledComponentCleanup.cleanComponents()` pass on the old id — e6c42eb. Capability/feature spec drain wording aligned in the same commit.

specs/020-revision-rollback/spec.md
Pasichniuk: Accepted — (a) was indeed a strict subset of (c) after re-creation stopped being a 404 condition. Applied your suggested FR-002 text verbatim in e6c42eb, and updated the "Restoring deleted resources" assumption in the same spec to reflect that only the (id-preserving) deployment path drains leftovers.

ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

PR #356, opened 2026-06-04
fix: scope cluster-local DNS matchPattern to namespace

### LINKED ISSUE #87: [Model Servings] Support text classification models

oleksii-donets:
### Applicable issues

- fixes #87

### Description of changes

For chained predictor + transformer inference deployments (reached today via
text-classification mode), the kube-dns egress in the generated
`CiliumNetworkPolicy` emitted a cluster-wide `matchPattern: *.svc.cluster.local`.
That granted the chained pair DNS reach across every namespace, even though the
transformer only ever needs to resolve its sibling predictor
(`<svc>-predictor.<namespace>.svc.cluster.local`) in its own namespace.

This change threads the policy's target namespace through `createKubeDnsEgress`
/ `toDnsList` and composes the pattern as `*.<namespace>.svc.cluster.local`,
narrowing DNS exfil exposure without changing same-namespace predictor
reachability. The wildcard branch (`allowedDomains=["*"]`) and the
non-chained / predictor-only shape are untouched.

- `CiliumNetworkPolicyCreator`: replace the `CLUSTER_LOCAL_DNS_PATTERN`
  constant with a suffix; build the pattern per namespace
- Tests: update the two pinned assertions and add a regression guard that
  varies the namespace across two policies
- Specs `022-transformer-cilium-policies` and `inference-deployments`: reflect
  the namespace-scoped pattern

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/pipeline/specification/CiliumNetworkPolicyCreator.java
```diff
@@ -193,7 +193,7 @@ private List<Dns> toDnsList(List<String> domains, boolean includeClusterLocal) {
         }
         if (includeClusterLocal) {
             Dns clusterLocal = new Dns();
-            clusterLocal.setMatchPattern(CLUSTER_LOCAL_DNS_PATTERN);
+            clusterLocal.setMatchPattern("*." + namespace + CLUSTER_LOCAL_DNS_SUFFIX);
```
Pasichniuk: Consider this refactoring to avoid string concatenation and magic string "*."
- CLUSTER_LOCAL_DNS_SUFFIX -> CLUSTER_LOCAL_DNS_PATTERN
- ".svc.cluster.local" -> "*.%s.svc.cluster.local"
- "*." + namespace + CLUSTER_LOCAL_DNS_SUFFIX -> CLUSTER_LOCAL_DNS_PATTERN.formatted(namespace)

oleksii-donets: done

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

Issue #162, opened 2026-02-19
[SPIKE] Add API to expose model metrics


oleksii-donets:
### Name and Version

0.16.0

### What is the problem this feature will solve?

Today, once a model deployment goes live, operators have **no first-class view into how it's actually performing**. We expose deployment state and pod status, but nothing about the workload running inside:

- No LLM-quality signals — Time-to-First-Token (TTFT), inter-token latency, tokens/sec, KV-cache utilization, queued requests.
- No resource-usage signals tied to a deployment — GPU utilization, VRAM pressure, CPU/memory headroom.
- No unified schema. vLLM, TGI, and NIM each expose a different metric vocabulary, so even when metrics exist, the UI and autoscaling logic can't consume them consistently.

### What is the feature you are proposing to solve the problem?

A **time-boxed spike** to design — and prove out with a lightweight PoC — a unified Model Metrics API for inference and NIM deployments, independent of the underlying engine.

#### Goals

1. **Pick a telemetry foundation.** Decide between Prometheus (pull-based, k8s-standard, metrics-only) and OpenTelemetry (unifies with the logs/traces we already export per `observability-and-logging`). Bias toward OTel given existing investment; validate with a short comparison.
2. **Define a unified metric schema** mapping engine-specific metrics to a common contract:
   - *Serving-quality*: TTFT, inter-token latency, tokens/s, queue depth, KV-cache utilization.
   - *Resource*: GPU utilization, GPU memory used/total, CPU, RAM, replica count.
   - *Operational*: error rate, p50/p95/p99 latency, scale events.
3. **Propose the API surface** — e.g. `GET /api/deployments/{id}/metrics` with time-range/granularity, plus a live-snapshot variant for the UI.
4. **PoC**: end-to-end flow for at least one vLLM-class inference deployment and one NIM deployment.

#### Why this matters — demo scenarios

- **"Is my new model fast enough?"** Right after deploying a vLLM model from HuggingFace, the operator opens the deployment page and sees live TTFT, tokens/sec, and KV-cache utilization — no external Grafana required.
- **"Why is this slow?"** A user reports a slow chat. The admin pivots from the DIAL request trace (already correlated by `traceparent`) into the deployment's metrics panel and sees a saturated KV cache plus queue buildup → root cause obvious.
[...]

PR #357, opened 2026-06-05
feat: deployment metrics proof-of-concept

Pasichniuk:
### Applicable issues

- related to #162

### Description of changes

Adds `GET /deployments/{id}/metrics` — a live, engine-neutral metrics snapshot for a deployment (PoC for a unified deployment-metrics API).

- **Unified snapshot endpoint** — `DeploymentController.getMetrics` returns resource metrics (replica counts + per-pod CPU/memory) for **any** deployment type, plus serving-quality metrics for `INFERENCE` deployments, in one engine-neutral schema with explicit units.
- **Auto engine detection + per-engine normalizers** — `EngineDetector` identifies the serving family (vLLM v0/v1, TGI, SGLang, KServe Python ModelServer) and dedicated normalizers map each engine's raw metrics onto the unified model (TTFT, inter-token/e2e latency, token throughput, queue depth, running requests, KV-cache utilization).
- **Hand-rolled Prometheus parser** — `PrometheusTextParser` + `HistogramSummaries` parse exposition format 0.0.4 and derive mean/p50/p95/p99/count from histograms, with **no new Gradle dependencies**.
- **Graceful partial results** — missing telemetry never fails the request: affected blocks return `null` with a reason in `availability`, response stays `200`; exactly one Ready pod is scraped (predictor for chained KServe deployments) and reported in `scrapedPod`.
- **Counter semantics & config** — counter-derived values are labelled lifetime aggregates (`window=lifetime`) with raw cumulative counters echoed in `rawCounters` for client-side rate derivation; request-triggered collection with a short response cache, toggleable via `app.metrics.scrape.*` (`400` when disabled).
- **Spec & docs** — adds the `deployment-metrics` capability spec (telemetry-foundation ADR, engine-availability matrix, sized follow-ups), the `023-deployment-metrics-api` speckit artifacts and OpenAPI contract, and `docs/configuration.md` rows.
- **Test coverage** — parser/normalizer/detector/service unit tests, H2 + K8s functional tests, and captured `/metrics` fixtures (real dev-cluster vLLM V1, plus vLLM v0/TGI/SGLang/KServe ModelServer).

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/VllmMetricsNormalizer.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        var serving = new ServingMetrics(
+                HistogramSummaries.summarize(index, TTFT),
+                interTokenLatency,
+                MetricSamples.lifetimeRate(index, PROMPT_TOKENS, now).orElse(null),
+                MetricSamples.lifetimeRate(index, GENERATION_TOKENS, now).orElse(null),
+                MetricSamples.asInteger(MetricSamples.sum(index, REQUESTS_WAITING)),
+                MetricSamples.asInteger(MetricSamples.sum(index, REQUESTS_RUNNING)),
+                MetricSamples.sum(index, KV_CACHE_USAGE_V0, KV_CACHE_USAGE_V1)
```
oleksii-donets: **Ratio gauge is summed across label sets, not averaged.** `kvCacheUsage` is a 0–1 ratio, but `MetricSamples.sum(...)` adds every matching series. If a pod exposes more than one series for the metric — vLLM V1 data-parallel with `engine="0"/"1"`, per-`model_name` labels, or (worst case) a build emitting both `vllm:gpu_cache_usage_perc` and `vllm:kv_cache_usage_perc` at once — the values are added (e.g. two engines at 0.4 → reported 0.8) and `clampRatio` then masks the inflation by capping at 1.0.

Ratio gauges should be averaged (or read from a single series), not summed. Counts like `runningRequests`/`queueDepth` are legitimately additive — this only applies to ratio gauges.

The same bug exists in `SglangMetricsNormalizer` for `sglang:token_usage` (the `MetricSamples.sum(index, TOKEN_USAGE).map(clampRatio)` call).

src/main/java/com/epam/aidial/deployment/manager/kubernetes/metrics/PodResourceUsageReader.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /** Sums CPU/memory across a pod's containers; {@code null} when the metrics carry no containers. */
+    private static PodResourceUsage toUsage(PodMetrics podMetrics, String podName) {
+        if (podMetrics == null || CollectionUtils.isEmpty(podMetrics.getContainers())) {
+            log.debug("No resource metrics for pod '{}'", podName);
+            return null;
+        }
+        double cpuMillicores = 0;
+        double memoryBytes = 0;
+        for (var container : podMetrics.getContainers()) {
```
oleksii-donets: **Summing all containers attributes sidecar usage to the deployment.** An inference pod is rarely single-container — KServe/KNative pods also run `queue-proxy`, often `istio-proxy`/Envoy, etc. Summing every container reports the whole-pod footprint, not the container the user actually configured. The reported CPU/memory then can't be compared to the user's `requests`/`limits` (which live on one container) and will look inflated/misleading in a per-deployment panel.

Consider filtering to the primary container (the deployment manager already knows the container name it generates — e.g. `kserve-container`) before summing, or returning a per-container breakdown so the consumer can decide. The same "primary container" notion will be needed for the GPU/DCGM follow-up.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
```diff
@@ -67,6 +67,9 @@
 @Slf4j
 public abstract class AbstractDeploymentManager<D extends Deployment, S> implements DeploymentManager<S> {
 
+    /** KServe stamps this label on every pod it owns to mark the component ({@code predictor} / {@code transformer}). */
+    private static final String KSERVE_COMPONENT_LABEL = "component";
```
oleksii-donets: **KServe-specific knowledge leaked into the abstract base.** A constant named `KSERVE_COMPONENT_LABEL` here, plus reading it in `toPodInfo` for *every* deployment type, contradicts this codebase's own documented rule — `specs/inference-deployments/spec.md`: *"no leak of the chained concept into the abstract base."* It's harmless at runtime for non-KServe pods (label absent → null), but the base class now knows a KServe vocabulary it shouldn't.

There's an established pattern to follow: `InferenceDeploymentManager` already overrides the `buildCiliumNetworkPolicy(...)` hook rather than putting KServe logic in the base. Do the same here — add a `protected String resolveComponent(Pod pod) { return null; }` hook on the base, call it from `toPodInfo`, and override it (with the `"component"` label constant) in `InferenceDeploymentManager`.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -0,0 +1,306 @@
...
+        }
+
+        var predictorPod = selectServingPod(readyPods);
+        var pod = predictorPod.getName();
+        // Assumption: the engine exposes Prometheus on its container/traffic port over plain HTTP
+        // (true for vLLM/TGI/SGLang/KServe ModelServer), and the transformer pod serves on the same
+        // port. Sidecar exporters on other ports (e.g. the KServe queue-proxy on 8012) are not read.
+        var port = deployment.getContainerPort() != null ? deployment.getContainerPort() : manager.getDefaultContainerPort();
+
+        var transformerFuture = selectTransformerPod(readyPods, predictorPod)
```
oleksii-donets: **Speculative transformer scrape runs for every inference deployment with a transformer pod, but is only consumed for `KSERVE_MODELSERVER`.** This future is started before engine detection. For a vLLM/TGI/SGLang deployment running on KServe with a transformer pod, the scrape executes (extra API-server pod-proxy round-trip + an occupied thread) and `buildScrapeContext` then discards the result — and the future is never cancelled. Worth gating the transformer scrape so it's not launched (or is cancelled) when the engine turns out not to be `KSERVE_MODELSERVER`.

src/main/java/com/epam/aidial/deployment/manager/model/metrics/UnifiedDeploymentMetrics.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Builder;
+
+import java.time.Instant;
+import java.util.Map;
+
+/**
+ * Layer-neutral live metrics snapshot for a model deployment — the unified, engine-neutral
+ * schema. Counter-derived values are lifetime aggregates (since engine process start),
+ * labelled via {@code window}; raw cumulative counters are echoed in {@code rawCounters}
+ * so clients can derive their own windowed rates.
```
oleksii-donets: **"clients can derive their own windowed rates" overclaims.** From a single snapshot's counters a client can only derive *short, forward-looking* rates over windows it observed itself — never arbitrary/historical windows ("last month"). It's also subject to counter resets on pod restart/redeploy (naive `now − then` goes negative) and to single-replica scope. Recommend narrowing the wording to "derive their own rates over the interval between their polls (subject to counter resets); arbitrary historical windows require the time-series backend (ADR Option C)."

This same phrasing recurs in `NormalizedEngineMetrics`, `DeploymentMetricsDto`, the controller `@Operation`, the `MetricSamples` Javadoc, and `specs/deployment-metrics/spec.md` — worth fixing consistently.

specs/deployment-metrics/spec.md
```diff
@@ -0,0 +1,162 @@
...
+#### Scenario: Raw counters echoed
+- **WHEN** a vLLM deployment is scraped
+- **THEN** `rawCounters` contains `prompt_tokens_total`, `generation_tokens_total`, and `request_success_total` as raw cumulative values
+
+#### Scenario: Percentile approximation
+- **WHEN** an engine exposes a native latency histogram
+- **THEN** p50/p95/p99 are interpolated from the cumulative `le` buckets, clamped at the last finite bucket boundary when the target rank falls into the `+Inf` bucket
+
+### Requirement: Request-triggered collection with bounded staleness
+Metric collection SHALL happen only in response to an API request — never via background polling (constitution Principle III). Repeated requests within a short interval are served from a per-deployment response cache with bounded staleness (default TTL 5 s) to limit kube API-server load. Degraded/partial snapshots are cached on the same terms as complete ones — there is no separate negative-caching path — so a transient failure (e.g. a momentary scrape timeout) may be served for up to the TTL before re-collection. Concurrent first requests for the same id are coalesced (single-flight) so only one collection runs per key per TTL.
```
oleksii-donets, line 78: **The cache and the "clients derive their own rates" contract interact, and that interaction isn't documented anywhere.** With this `expireAfterWrite` cache, a client polling faster than the TTL gets byte-identical snapshots (same `collectedAt`, same counters). If it differences on its own wall-clock → `Δcounter = 0` over `Δt > 0` → derived rate `0` (silently wrong); if it differences on `collectedAt` → `0/0` within the TTL, which it must detect and skip. Net effect: **effective rate resolution is capped at the cache TTL, not the client's poll interval** — and naive differencing under-counts.

`collectedAt` is the latent mitigation, but no doc tells clients to key off it and dedupe identical snapshots. Suggest adding a note cross-referencing this requirement and the raw-counters requirement (line ~63), e.g.:

> Rate derivation must use `collectedAt` as the time base, not the client's request time. Consecutive responses within the cache TTL return the same snapshot and represent a single data point; effective rate resolution is bounded by the cache TTL.

(Also worth confirming the deferred `rate-window.*` flag — disabled/not implemented — isn't implied as available in `docs/configuration.md`.)

src/main/java/com/epam/aidial/deployment/manager/web/controller/DeploymentController.java
```diff
@@ -217,6 +222,22 @@ public SseEmitter subscribeToLogs(
...
 
+    @GetMapping(path = "{id}/metrics",
+            produces = MimeTypeUtils.APPLICATION_JSON_VALUE)
+    @Operation(summary = "Get a live metrics snapshot for a deployment",
+            description = "Resource metrics (replica counts and per-pod CPU/memory) are reported for any deployment "
+                    + "type; serving-quality metrics are additionally scraped for INFERENCE deployments. Missing "
+                    + "telemetry never fails the request: affected blocks are null with the reason recorded in "
+                    + "'availability' and the response is still 200. Counter-derived values are lifetime aggregates "
+                    + "(window=lifetime); raw cumulative counters are echoed in 'rawCounters' for client-side rate "
+                    + "derivation. Exactly one Ready pod is scraped for serving metrics and named in 'scrapedPod'.")
```
oleksii-donets: **"Exactly one Ready pod is scraped" is inaccurate for chained KServe.** A chained `KSERVE_MODELSERVER` deployment scrapes **two** pods (predictor + transformer); only the predictor is named in `scrapedPod`. Reword to match the capability spec, e.g. "the Ready predictor pod is scraped and named in `scrapedPod`; a chained KServe deployment additionally scrapes the transformer pod."

src/main/java/com/epam/aidial/deployment/manager/web/dto/metrics/DistributionSummaryDto.java
```diff
@@ -0,0 +1,8 @@
+package com.epam.aidial.deployment.manager.web.dto.metrics;
+
+/**
+ * Latency distribution summary. Percentiles are approximated from cumulative Prometheus
+ * histogram buckets in a single snapshot (lifetime window).
+ */
+public record DistributionSummaryDto(Double mean, Double p50, Double p95, Double p99, long count) {
```
oleksii-donets: **"Explicit units" is claimed but not expressed in the schema.** The PR/spec emphasize an engine-neutral schema "with explicit units", but these latency fields (`mean`/`p50`/`p95`/`p99`) carry no unit marker — they're seconds, discoverable only from the spec's matrix, not the API contract. Contrast with `cpuMillicores`/`memoryBytes`, which are self-describing. Consider documenting the unit on the DTO/OpenAPI (the values are seconds) so clients don't have to infer it.

src/main/java/com/epam/aidial/deployment/manager/model/metrics/PodResourceUsage.java
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Per-pod resource consumption. GPU fields are present in the contract but always {@code null}
+ * in the PoC — they require the DCGM exporter cluster prerequisite (follow-up).
+ */
+public record PodResourceUsage(
+        String name,
+        Double cpuMillicores,
+        Double memoryBytes,
+        Double gpuUtilization,
+        Double gpuMemoryUsedBytes
```
oleksii-donets: **Field naming isn't unified.** Two inconsistencies:
- `memoryBytes` vs `gpuMemoryUsedBytes` — the "Used" qualifier is on one but not the other. Both are "bytes currently used", so align them (`memoryBytes`/`gpuMemoryBytes`, or `memoryUsedBytes`/`gpuMemoryUsedBytes`).
- `gpuUtilization` is the only field without a unit suffix (it's a bare ratio). Defensible since it's a ratio, not an amount, but it reads oddly next to four unit-suffixed siblings; `gpuUtilizationRatio` would make it explicit.

Note this name is load-bearing in `ResourceMetricsDto.PodResourceUsageDto` and the implicit MapStruct mapping (matches by property name) plus the OpenAPI contract — so rename now while GPU fields are always `null`, before any client depends on them.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -0,0 +1,306 @@
...
+    private final PrometheusTextParser prometheusTextParser;
+    private final EngineDetector engineDetector;
+    private final List<EngineMetricsNormalizer> normalizers;
+    private final PodResourceUsageReader podResourceUsageReader;
+    private final MetricsScrapeProperties properties;
+
+    @Cacheable(cacheNames = MetricsCachingConfig.DEPLOYMENT_METRICS_CACHE_NAME, sync = true)
+    public UnifiedDeploymentMetrics getSnapshot(String id) {
+        if (!properties.isEnabled()) {
+            throw new IllegalArgumentException("Metrics collection is disabled by configuration (app.metrics.scrape.enabled=false)");
```
oleksii-donets: **Disabled-by-config path is reported as an uncaught error.** This throws `IllegalArgumentException`, which `DefaultExceptionHandler.handleIllegalArgumentError` maps to 400 (correct) but routes through `logUncaught(ex)` — so a normal, intended config state produces error-level "uncaught" log noise on every call. Consider a dedicated exception (or a handler that doesn't log-as-uncaught) for this case. Minor: "feature disabled" is arguably 403/503 rather than 400, though 400 matches the documented contract.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    /**
+     * Chooses which Ready pod to scrape for serving metrics. A KServe InferenceService with a
+     * transformer carries the engine on its {@code predictor} pod while the {@code transformer} pod
+     * exposes no engine metrics; both share the InferenceService label, so the pod listing returns
+     * both in an undefined order. Prefer the predictor when present; otherwise (KNative/raw
+     * inference, single-pod, or no component labels) fall back to the first Ready pod.
+     */
+    private static PodInfo selectServingPod(List<PodInfo> readyPods) {
```
oleksii-donets: **Serving metrics come from a single replica while resource metrics cover all pods — inconsistent granularity.** This scrapes one Ready predictor, so per-pod gauges (`queueDepth`, `runningRequests`, `kvCacheUsage`) reflect only that replica, whereas `resources` aggregates all pods. For `replicas > 1` the serving numbers can mislead. Multi-pod aggregation is explicitly out of scope (per the spec), so this is a documentation/caveat note rather than a fix — but the per-deployment framing of those gauges isn't caveated in the response/contract.

src/main/java/com/epam/aidial/deployment/manager/web/mapper/DeploymentMetricsDtoMapper.java
```diff
@@ -0,0 +1,60 @@
...
+                toDto(serving.interTokenLatency()),
+                new ServingMetricsDto.TokensPerSecondDto(serving.promptTokensPerSecond(), serving.generationTokensPerSecond()),
+                serving.queueDepth(),
+                serving.runningRequests(),
+                serving.kvCacheUsage(),
+                toDto(serving.requestLatency()),
+                serving.requestsPerSecond());
+    }
+
+    default ResourceMetricsDto toDto(ResourceMetrics resources) {
```
oleksii-donets: **Cleanup (optional):** this `default` method (and the `ServingMetricsDto` one above) can be pure-annotation, consistent with the rest of this mapper. MapStruct 1.6.3 (this repo) supports nested-target paths into records:

```java
@Mapping(target = "replicas.total", source = "replicasTotal")
@Mapping(target = "replicas.ready", source = "replicasReady")
ResourceMetricsDto toDto(ResourceMetrics resources);

@Mapping(target = "tokensPerSecond.prompt",     source = "promptTokensPerSecond")
@Mapping(target = "tokensPerSecond.generation", source = "generationTokensPerSecond")
ServingMetricsDto toDto(ServingMetrics serving);
```

`pods` auto-maps via the existing element method; null-handling is preserved. Purely a consistency/brevity nit — the explicit `default` form is also fine.

src/main/java/com/epam/aidial/deployment/manager/model/metrics/BlockAvailability.java
```diff
@@ -0,0 +1,13 @@
+package com.epam.aidial.deployment.manager.model.metrics;
+
+/**
+ * Per-block availability marker; {@code reason} is human-readable and present only when degraded.
+ */
+public record BlockAvailability(boolean available, String reason) {
```
oleksii-donets: **Naming nit: `BlockAvailability` is a bit vague / mismatched.** "block" is domain vocabulary (the spec defines a per-block `{available, reason}` marker, and the availability map keys are the blocks), so it's not invented — but:

- The type itself is a generic `(boolean available, String reason)` pair and knows nothing about "blocks"; the block identity lives in the *map key* (`Map<String, BlockAvailability>`), not the value. So the `Block` prefix describes the usage context, not the type.
- It's inconsistent with its own wire type, `MetricsAvailabilityDto` — "Block" is dropped and replaced with "Metrics", so the concept is already named two different ways across layers.

What it actually is is an availability status. Consider `AvailabilityStatus` (most accurate to the shape) or `MetricAvailability` (aligns model ↔ DTO). Minor, naming-only — the current name is traceable to the spec, just not the clearest and not consistent with the DTO.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -0,0 +1,306 @@
...
+        return readyPods.stream()
+                .filter(pod -> COMPONENT_PREDICTOR.equalsIgnoreCase(pod.getComponent()))
+                .findFirst()
```
oleksii-donets, line 129: **Follow-up to the single-replica note above — `findFirst()` over an "undefined order" list makes replica selection non-deterministic, and that breaks the `rawCounters` rate contract.** With N≥2 predictor replicas this still returns a valid pod, but:

- The Javadoc itself says the listing order is undefined, so across cache-TTL-separated polls a *different* replica may be scraped each time. Gauges then jump (e.g. `runningRequests`, `kvCacheUsage`) for reasons unrelated to real load.
- Worse: `rawCounters` are cumulative *since that replica's process start*. When the sampled replica switches between two polls, a client differencing a counter sees it **go backwards** — indistinguishable from a counter reset. So the documented "derive your own rates from rawCounters" fails on every replica switch, not just on pod restart.

Minimum mitigation: make selection deterministic (e.g. sort predictors by name before `findFirst`) so the same replica is sampled across polls. Proper fix is multi-replica aggregation (sum counters, merge histogram buckets), which the spec lists as out-of-scope — so at minimum the contract should state serving metrics are single-replica (and name which replica), so clients don't treat them as deployment-wide or feed them into naive rate math.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -0,0 +1,306 @@
...
+                .orElseGet(readyPods::getFirst);
```
oleksii-donets: **The fallback target is the weak point — "if no predictor, take the first Ready pod" conflates two different situations.** The `filter → findFirst → orElse(first)` shape is fine; the issue is what "first" means here:

1. **No `component` labels at all** (KNative/raw inference, single-pod, NIM) → first Ready pod is correct.
2. **Component-labelled but no predictor Ready right now** → "first Ready pod" can be the **transformer**, which exposes no engine metrics.

Case 2 is reachable: chained KServe with the predictor still starting and the transformer already Ready → `readyPods = [transformer]`, predictor filter empty, `getFirst()` returns the transformer. It then gets scraped, detection finds no `vllm:`/`tgi_`/`sglang:` prefix and no `request_predict_seconds` marker (a transformer declares `request_preprocess_seconds`/`request_postprocess_seconds`, not predict) → engine `UNKNOWN`. So the response reports **"engine not recognized"** with `scrapedPod` = the transformer, when the truthful reason is **"no Ready predictor yet."** Misleading diagnostics, and it scraped a pod that was never going to carry engine metrics.

Suggest distinguishing the two — only fall back to "first" when there are no component labels at all; otherwise treat "no Ready predictor" as its own degradation:

```java
boolean componentLabelled = readyPods.stream().anyMatch(p -> p.getComponent() != null);
return readyPods.stream()
        .filter(p -> COMPONENT_PREDICTOR.equalsIgnoreCase(p.getComponent()))
        .findFirst()
        .orElseGet(() -> componentLabelled ? null /* → "no ready predictor" reason */
                                           : readyPods.getFirst());
```

(For what it's worth, `getFirst()` itself is safe — the caller already guards `readyPods` non-empty, so no `NoSuchElementException`.)

src/main/java/com/epam/aidial/deployment/manager/kubernetes/metrics/PodResourceUsageReader.java
```diff
@@ -0,0 +1,100 @@
...
+     * then filters to {@code podNames} client-side — one API round-trip regardless of replica
+     * count, instead of one per pod. Pods without metrics are simply absent from the result.
+     */
+    public List<PodResourceUsage> readAll(String namespace, Collection<String> podNames) {
+        if (CollectionUtils.isEmpty(podNames)) {
+            return List.of();
+        }
+        Set<String> wanted = Set.copyOf(podNames);
+        try {
+            var podMetricsList = client.top().pods().metrics(namespace);
```
oleksii-donets, line 71: **This fetches every pod's usage in the namespace and discards all but `wanted` — scope it to the deployment instead.** The "one round-trip regardless of replica count" goal is right, but in a shared model namespace `metrics(namespace)` returns usage for *every* pod each call and keeps ~1–3, so transfer + deserialization + metrics-server load scale with namespace size, not deployment size — repeated per deployment (bounded only by the 5 s cache).

`metrics.k8s.io` honors `labelSelector`, and the deployment's pods already share a service label (the one the pod listing filters on). A label-scoped query gives both properties at once — one round-trip *and* only this deployment's pods:

```java
client.resources(PodMetrics.class, PodMetricsList.class)
      .inNamespace(namespace)
      .withLabels(serviceLabelSelector)
      .list();
```

(Use the generic typed-resource form — `top().pods()` likely doesn't expose label filtering in this Fabric8 version, which is probably why namespace-wide was chosen. Worth confirming the label selector reaches metrics-server before committing.)

Per-pod calls (`metrics(namespace, podName)`) are *not* the alternative — that's N round-trips, exactly what this avoids. If label-scoping turns out unavailable, namespace-wide is a defensible fallback and the client-side filter is correct — just flagging the scaling cost.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -0,0 +1,306 @@
...
+            log.debug("No service for deployment '{}' yet ({}); treating as no pods", id, e.getMessage());
+            return new DeploymentManager.PodInstances(List.of(), List.of());
+        }
+    }
+
+    /**
+     * Serving-quality metrics are scraped only for INFERENCE deployments; every other type reports
+     * the serving/operational blocks as unavailable (resource metrics still apply to all types).
+     */
+    private EngineScrapeResult collectServingMetrics(Deployment deployment, DeploymentManager<?> manager,
```
oleksii-donets: **Design: extract serving-metrics collection into a per-type `supports()`-based collaborator now — NIM is a second consumer, so this isn't premature.** This service currently bundles three responsibilities: top-level orchestration/caching, resource metrics (all types), and an inference-specific cluster (predictor/transformer selection, scrape orchestration + transformer join, engine detect/normalize) gated on `instanceof InferenceDeployment`. The engine *vocabulary* is already nicely extracted (`EngineDetector`, normalizers, `HistogramSummaries`), but the inference-specific orchestration is inline.

NIM is already a first-class type that exposes its own Prometheus metrics and today gets resource-only. When NIM serving metrics land, the inline approach grows an `instanceof` ladder (`InferenceDeployment` → `NimDeployment` → …). Establishing the seam now avoids that:

```java
interface ServingMetricsCollector {
    boolean supports(Deployment deployment);                       // mirrors EngineMetricsNormalizer.supports()
    NormalizedEngineMetrics collect(EngineScrapeTarget target);    // pods, port, namespace
}
```

`DeploymentMetricsService` then orchestrates: resolve deployment + pods → **always** collect resources (stays shared, not per-type) → pick the matching `ServingMetricsCollector` (or none) → assemble + cache. Adding NIM = add `NimServingMetricsCollector`, no change to the orchestrator. This also resolves the layering note (#3 on `AbstractDeploymentManager`): the predictor/transformer `component`-label knowledge moves into the inference collector instead of living in generic code.

Guardrails so this doesn't become premature abstraction:
- Shape the interface around what the orchestrator needs (supports + produce `NormalizedEngineMetrics` from pods); keep port/path/detection private to each collector so the interface is stable regardless of NIM internals.
- Build the seam + the inference impl now; **don't** pre-implement NIM — let NIM validate the seam when its metrics actually land.
- Confirm the abstraction is carved where NIM actually differs (NIM likely reuses vLLM-ish vocab on a different port — the create request already carries a distinct `containerPort`/`containerGrpcPort`).

Maintainability/cohesion improvement, not a defect — the current code works.

oleksii-donets (commented):
Review of the deployment-metrics PoC. Solid, well-tested work overall — the Prometheus parser and histogram percentile math are correct, and graceful degradation is mostly honored. Comments are grouped into: a few behavioral issues (ratio gauges summed not averaged, per-pod usage summing sidecars, namespace-wide metrics fetch), a layering/SRP thread (KServe specifics leaking into generic code; extracting a per-type serving-metrics collector now that NIM is a second consumer), and a cluster of contract/doc clarifications around the "lifetime" counters, caching, and client-side rate derivation. Most are non-blocking; details inline. Nothing here blocks merge as a PoC — flagging the items worth resolving before this hardens into a stable contract (the GPU field naming and the rawCounters/lifetime semantics especially, since those are wire-contract decisions that get expensive to change later).

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/VllmMetricsNormalizer.java
Pasichniuk: Fixed — ratio gauges now use `MetricSamples.avg()` (`kvCacheUsage`/`token_usage`); additive counts stay summed.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/metrics/PodResourceUsageReader.java
Pasichniuk: Fixed — usage is now attributed to the primary workload container (via `PodInfo.mainContainerName` / `getContainerName`), excluding injected sidecars.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
Pasichniuk: Fixed — dropped the constant from the base and added a `resolveComponent(Pod)` hook, overridden in `InferenceDeploymentManager` (mirrors the `buildCiliumNetworkPolicy` pattern).

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Fixed — the transformer scrape is now gated on `KSERVE_MODELSERVER` after engine detection (serialized), so no wasted scrape and no leaked future.

src/main/java/com/epam/aidial/deployment/manager/model/metrics/UnifiedDeploymentMetrics.java
Pasichniuk: Fixed — narrowed to rates over the polling interval (subject to counter resets); arbitrary historical windows require the TSDB. Applied consistently across the listed spots.

specs/deployment-metrics/spec.md
Pasichniuk: Documented — rate derivation must key off `collectedAt` and dedupe identical snapshots; effective resolution is bounded by the cache TTL. Added to the spec and DTO/model Javadoc.

src/main/java/com/epam/aidial/deployment/manager/web/controller/DeploymentController.java
Pasichniuk: Fixed — reworded to "Ready predictor pod … a chained KServe Python ModelServer deployment additionally scrapes the transformer pod" to match the spec.

src/main/java/com/epam/aidial/deployment/manager/web/dto/metrics/DistributionSummaryDto.java
Pasichniuk: Fixed — added `@Schema(description = "… in seconds")` to the `mean`/`p50`/`p95`/`p99` fields (and noted it in the Javadoc).

src/main/java/com/epam/aidial/deployment/manager/model/metrics/PodResourceUsage.java
Pasichniuk: Renamed `gpuMemoryUsedBytes` → `gpuMemoryBytes` across model/DTO/spec/contract while GPU fields are still null; kept `gpuUtilization` (a bare ratio, consistent with `kvCacheUsage`).

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Fixed — added a dedicated `MetricsCollectionDisabledException` → 400 with a handler that skips `logUncaught`. Minor note: `logUncaught` is DEBUG-only, so it was debug-level noise rather than error-level, but the "uncaught" semantic was indeed wrong.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Caveated — the spec and `ServingMetricsDto`/`OperationalMetricsDto` Javadoc now state serving/operational metrics reflect the single scraped replica (`scrapedPod`), not a cross-replica aggregate. Multi-pod aggregation stays out of scope.

src/main/java/com/epam/aidial/deployment/manager/web/mapper/DeploymentMetricsDtoMapper.java
Pasichniuk: Done — both `default` methods are now pure-annotation with nested-target `@Mapping` (`tokensPerSecond.*`, `replicas.*`); verified the generated impl maps correctly and preserves null-handling.

src/main/java/com/epam/aidial/deployment/manager/model/metrics/BlockAvailability.java
Pasichniuk: Renamed to `AvailabilityStatus` (the wire type `MetricsAvailabilityDto` stays).

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Fixed — selection is now deterministic (lowest pod name) for both predictor and transformer, so the same replica is sampled across polls; also documented serving metrics as single-replica.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Fixed — fall back to the first Ready pod only when there are no component labels; a component-labelled deployment with no Ready predictor now degrades with a dedicated "no ready predictor" reason instead of scraping the transformer.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/metrics/PodResourceUsageReader.java
Pasichniuk: Kept namespace-wide for now (one round-trip, client-side filter is correct) and added a comment flagging the scaling cost; label-scoped query noted as a follow-up pending confirmation that metrics-server honors `labelSelector`.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Done — extracted a `supports()`-based `ServingMetricsCollector` with an `InferenceServingMetricsCollector` impl; `DeploymentMetricsService` is now a type-agnostic orchestrator, so NIM later = a new collector. No behavior change (all existing tests pass).

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
oleksii-donets: Following up now that this moved into `InferenceServingMetricsCollector.selectServingPod`: the two-case split is right, and the "labelled but no Ready predictor → `null` → `REASON_NO_READY_PREDICTOR`" branch is exactly the truthful degradation I was after. 👍

One leftover, though: the `componentLabelled ? null : readyPods.getFirst()` fallback to the first Ready pod is effectively unreachable for this collector. `supports()` is `instanceof InferenceDeployment` (KServe-only), and KServe always stamps `component=predictor`/`transformer` on its pods (both KNative and RawDeployment modes), so `componentLabelled` is always `true` here. The javadoc justifies the fallback with "KNative/raw inference, single-pod, NIM" — but NIM is a separate type that gets its own `ServingMetricsCollector` (the whole reason we extracted the seam), and single-pod KServe is still `component`-labelled, so none of those cases actually flow through this branch.

Not a correctness problem — if it ever did fire, `EngineDetector` would return `UNKNOWN` → `REASON_UNKNOWN_ENGINE`, so it can't misreport metrics. But it's dead defensiveness that quietly reintroduces the "scrape whichever pod sorts first" non-determinism the rest of this method works to avoid. Suggest simplifying to just `return null` when no predictor is found (the caller already degrades with the truthful reason), or at minimum dropping the misleading NIM/single-pod rationale from the javadoc.

Pasichniuk: Confirmed and simplified — `selectServingPod` now returns the deterministic predictor or `null` (caller degrades with the truthful "no ready predictor" reason), and the misleading NIM/single-pod rationale is dropped from the javadoc.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #367, opened 2026-06-16
fix: bind image entrypoint/cmd from container config blob

### LINKED ISSUE #352: Upgrade Spring Boot version to 4.x
### Name and Version

ai-dial-admin-deployment-manager-backend 0.17.0

According to [Spring Boot Support page](https://spring.io/projects/spring-boot#support), version 3.5.x reaches end of OSS support this month. Thus, to continue receive dependency updates with vulnerability fixes upgrade to 4.x becomes mandatory.

Upgrade Spring Boot to 4.x.

_No response_

oleksii-donets:
### Applicable issues

- fixes #352

### Description of changes

Follow-up to the Spring Boot 4 / Jackson 3 (`tools.jackson`) upgrade.

`ContainerConfigurationTemplateDto.config` was declared `final`, so Lombok `@Data` generated no setter. Combined with the project mapper disabling `DETECT_PARAMETER_NAMES` (no constructor binding), Jackson had no mutator for the `config` property and silently kept the inline-initialized empty object — so `DockerRegistryClient.getEntrypoint(...)` always returned an empty entrypoint/cmd.

- Drop `final` on `config` so `@Data` generates the setter and Jackson binds the nested container config.
- Switch `DockerRegistryClientTest` to the real project `ObjectMapper` (`JsonMapperConfiguration.createJsonMapper()`) instead of mocking `readValue`, so the actual Jackson binding is exercised. Verified the test fails against the old `final` field, locking in the regression.

### Checklist

### DIFF
--- a/src/main/java/com/epam/aidial/deployment/manager/docker/dto/ContainerConfigurationTemplateDto.java
+++ b/src/main/java/com/epam/aidial/deployment/manager/docker/dto/ContainerConfigurationTemplateDto.java
@@ -13,7 +13,7 @@ import java.util.List;
 )
 public class ContainerConfigurationTemplateDto {
 
-    private final ConfigurationObjectTemplate config = new ConfigurationObjectTemplate();
+    private ConfigurationObjectTemplate config = new ConfigurationObjectTemplate();
 
     @Data
     public static class ConfigurationObjectTemplate {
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

PR #377, opened 2026-06-29
feat: expose model serving capability API

dearsimanovich-epam:
### Applicable issues

- fixes #

### Description of changes

Exposes what a deployed inference model supports so the frontend can route it to the right surface: **text-generation → chat completion endpoint**, **text-classification → MCP toolset**, none → neither.

- **Detection** (`InferenceTaskDetector`): recognises text-generation via HuggingFace `pipeline_tag == "text-generation"` or a causal-LM architecture (`*ForCausalLM` / `*ForConditionalGeneration` / `*LMHeadModel`), reusing metadata already fetched. Precedence: if a model matches both signals, text-classification wins (preserves the existing transformer-chaining contract).
- **Persistence**: new `inference_deployment.inference_task` column (Flyway `V1.59`, H2/POSTGRES/MS_SQL_SERVER), `VARCHAR(32) DEFAULT 'NONE' NOT NULL` (the `DEFAULT` backfills existing rows in one step); `_aud` audit column nullable at the Envers default length. Computed at create and on source-changing update via `DeploymentManager#enrichBeforePersist`. `PersistenceDeploymentMapper` coalesces null → `NONE` at the persistence boundary so no save path can write null.
- **API**: read-only `inferenceTask` (`TEXT_GENERATION` / `TEXT_CLASSIFICATION` / `NONE`) on the inference deployment response (single + list); absent from the create/update request (server-computed only). Excluded from config export/import.
- Deploy-time manifest path (transformer chaining) unchanged — `TEXT_GENERATION` is predictor-only, like `NONE`.

Specs: `specs/024-model-serving-capability/` and updates to `specs/inference-deployments/spec.md`.

> Follow-up review comments are addressed in #379.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/resources/db/migration/H2/V1.59__AddInferenceTaskColumn.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE inference_deployment ADD COLUMN inference_task VARCHAR(255);
```
oleksii-donets: Nit (consistency): `VARCHAR(255)` is oversized for the main table here. Every enum-as-string column on the non-audit tables uses `VARCHAR(32)` — including the sibling `inference_deployment.model_format VARCHAR(32)` added in V1.38, plus `deployment.status`, `component_removal.type`, `build_status`, `transport_type`. The longest value (`TEXT_CLASSIFICATION`) is 19 chars, so `VARCHAR(32)` fits with headroom and matches convention.

```suggestion
ALTER TABLE inference_deployment ADD COLUMN inference_task VARCHAR(32);
```

The `_aud` column below is correctly `VARCHAR(255)` — that matches the audit-table convention from `V1.55__CreateAuditTables.sql` (Envers default length), so leave it as-is. Same `VARCHAR(32)`/`VARCHAR(255)` split applies to the POSTGRES and MS_SQL_SERVER copies.

Not a functional issue (Postgres/H2 don't allocate by declared length, and `ddl-auto: validate` doesn't check `VARCHAR` length) — purely convention/readability.

src/test/java/com/epam/aidial/deployment/manager/functional/config/FunctionalTestConfiguration.java
```diff
@@ -93,6 +97,24 @@ public SecurityClaimsExtractor securityClaimsExtractor() {
...
+    /**
+     * Stub HuggingFace client so inference-deployment create/update (which now runs task detection
+     * via {@code InferenceDeploymentManager#enrichBeforePersist}) does not make real HTTP calls in
+     * functional tests. Returns metadata that resolves to {@code InferenceTask.NONE}. Marked
+     * {@link Primary} to win over the scanned real client; tests that need a specific task can
+     * re-stub this mock.
+     */
+    @Bean
+    @Primary
+    public HuggingFaceClient testHuggingFaceClient() {
```
oleksii-donets: `@Primary` isn't needed here — and it's the only mock in this class that uses it. The reason it became necessary is the method name: `testHuggingFaceClient()` registers a *second* bean of type `HuggingFaceClient` (distinct from the scanned `huggingFaceClient`), creating the injection ambiguity that `@Primary` then resolves.

Compare `dockerRegistryClient()` a few methods up (`DockerRegistryClient` is `@Service` and its package is scanned too): it mocks the real bean **without** `@Primary` simply because its method name matches the scanned bean name — an explicit `@Bean` overrides a component-scanned bean of the same name (no `allow-bean-definition-overriding` required; the green suite proves it).

So just rename to match the scanned bean and drop `@Primary`:

```suggestion
    @Bean
    public HuggingFaceClient huggingFaceClient() {
```

This mirrors the existing `dockerRegistryClient` mock in this same class rather than introducing a one-off `@Primary`, and the rest of the method body is unchanged.

oleksii-donets (commented):
Solid, well-scoped change — detection precedence, manifest preservation (FR-009), null→`NONE` coalescing, `@JsonIgnore` for export/import, and the multi-vendor migrations all look correct, and spec maintenance is thorough.

Two minor inline nits below (migration column size; dropping `@Primary` in the test stub).

src/main/resources/db/migration/H2/V1.59__AddInferenceTaskColumn.sql
dearsimanovich-epam: Done in a304bf9a — main-table column is now `VARCHAR(32)` across H2/POSTGRES/MS_SQL_SERVER; the `_aud` column stays `VARCHAR(255)` per the Envers convention. Thanks for the convention pointer.

src/test/java/com/epam/aidial/deployment/manager/functional/config/FunctionalTestConfiguration.java
dearsimanovich-epam: Done in a304bf9a — renamed to `huggingFaceClient()` so it overrides the scanned bean by name (mirrors `dockerRegistryClient()`) and dropped `@Primary`. Functional suite stays green.

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/resources/db/migration/H2/V1.60__BackfillInferenceTask.sql
```diff
@@ -0,0 +1,5 @@
+-- Backfill existing inference deployments created before the inference_task column existed.
+-- Sets them to NONE (the conservative default the API already reports for null); the real task is
+-- re-detected on the deployment's next deploy/update. Audit (_aud) rows are left as-is to preserve
+-- historical revision state.
+UPDATE inference_deployment SET inference_task = 'NONE' WHERE inference_task IS NULL;
```
oleksii-donets: A separate backfill migration isn't needed here — V1.59 is introduced in this same PR and hasn't been released yet, so you can just modify it to add the column with a default in one step, instead of add-nullable-then-UPDATE:

```sql
-- V1.59 (H2 / POSTGRES)
ALTER TABLE inference_deployment ADD COLUMN inference_task VARCHAR(32) DEFAULT 'NONE' NOT NULL;
```
```sql
-- V1.59 (MS_SQL_SERVER)
ALTER TABLE inference_deployment ADD inference_task VARCHAR(32) NOT NULL DEFAULT 'NONE';
```

`ADD COLUMN … DEFAULT 'NONE'` populates existing rows with `NONE` atomically on all three vendors, so this whole V1.60 `UPDATE … WHERE inference_task IS NULL` becomes unnecessary. `NOT NULL` also makes the data self-consistent (and lets you later drop the null→`NONE` coalescing in `DeploymentDtoMapper`). Editing V1.59 in place is fine — the migration-immutability rule applies to migrations already released to `development`/`main`, not one added in the same open PR. Keep the `_aud` column nullable as you have it (historical revisions legitimately have no value).

⚠️ One prerequisite before going `NOT NULL`: the persist path must never write null. Today `enrichBeforePersist` leaves `inferenceTask` null when the source isn't a `HuggingFaceSource` (`InferenceDeploymentManagerTest#enrichBeforePersist_shouldLeaveTaskUnset_whenSourceIsNotHuggingFace` asserts exactly that). Hibernate includes the column in its INSERT with an explicit `NULL`, and a column `DEFAULT` does **not** apply when the column is named with a null value — so that insert would fail the `NOT NULL` constraint. Before relying on `NOT NULL`, either default the domain/entity field to `NONE`, or have `enrichBeforePersist` set `NONE` for non-HuggingFace/absent sources. (If inference deployments are always HuggingFace-sourced and detection always returns one of the three values, that null path may be unreachable in practice — worth confirming.)

If you'd prefer to keep the column nullable, V1.60 as-is is also fine — the two approaches are equivalent for the API contract. Just flagging that one migration can do both jobs.

oleksii-donets (commented):
One follow-up on the new V1.60 backfill — it can be folded into V1.59 (add the column `DEFAULT 'NONE' NOT NULL` in one step) rather than shipping a second migration. Details + a NOT NULL prerequisite inline.

src/main/resources/db/migration/H2/V1.60__BackfillInferenceTask.sql
dearsimanovich-epam: Done in 10cfe64c — folded the backfill into V1.59 and deleted V1.60. The main column is now `VARCHAR(32) DEFAULT 'NONE' NOT NULL` (the `DEFAULT` backfills existing rows atomically on all three vendors); `_aud` stays nullable.

For the NOT NULL prerequisite you flagged, I guarded the single persistence choke point instead of relying on the create path alone: `PersistenceDeploymentMapper` now coalesces a null `inferenceTask` → `NONE` when mapping to the entity. Both `DeploymentRepository.save()` and `update()` build the entity via `toEntity()`, so create, update, **and rollback from a pre-feature audit revision** (where `_aud` is legitimately null) are all covered — no null can reach the NOT NULL column. Entity column marked `nullable = false`; full H2 functional suite (which inserts into the column) stays green.

Kept the domain field nullable + the API null→`NONE` coalescing as defensive depth; can drop the latter as a follow-up now that the column is NOT NULL.

ai-dial-actions:
<h1>Dependency Review</h1>
None

oleksii-donets (approved):

CLAUDE.md
```diff
@@ -70,3 +70,10 @@ Where `<slug>` is the directory name under `specs/` that the feature belongs to
...
 2. Read the `**Capability**:` line; for each listed slug, open `specs/<slug>/spec.md` and update the affected Requirement(s)/Scenario(s) to match the shipped behaviour. Add a one-line `Implemented via NNN-<feature>` cross-reference near the changed Requirement, and remove or upgrade any `Pending: NNN-<feature>` note for this feature.
 3. If the field is `N/A — creates new capability <slug>`, scaffold `specs/<slug>/spec.md` from the house style (use `specs/api-conventions/spec.md` as a model) and add a row to `specs/README.md`.
+
+## Active Technologies
+- Java 21 + Spring Boot 4.0.6 (Framework 7), Hibernate ORM 7 + Envers, MapStruct 1.6.3, Flyway, Fabric8/KServe client, Lombok (024-model-serving-capability)
+- Relational, multi-vendor (H2 / PostgreSQL / SQL Server) via Flyway migrations; `inference_deployment` table (024-model-serving-capability)
+
+## Recent Changes
+- 024-model-serving-capability: Added Java 21 + Spring Boot 4.0.6 (Framework 7), Hibernate ORM 7 + Envers, MapStruct 1.6.3, Flyway, Fabric8/KServe client, Lombok
```
Pasichniuk, line 79: Redundant change. Recent changes are false claims

src/main/java/com/epam/aidial/deployment/manager/web/dto/deployment/InferenceDeploymentDto.java
```diff
@@ -18,4 +19,12 @@ public class InferenceDeploymentDto extends DeploymentDto {
     private String modelFormat;
     @NotNull @Valid
     private InferenceDeploymentSourceDto source;
+
+    /**
+     * System-computed serving capability (read-only). The frontend maps it to a consumption
+     * surface: {@code TEXT_GENERATION} → chat completion, {@code TEXT_CLASSIFICATION} → MCP
+     * toolset, {@code NONE} → neither. Never null in responses (a missing detected value is
+     * reported as {@code NONE}).
+     */
```
Pasichniuk, line 28: Redundant info about frontend. Claims it's never null, but doesn't have NotNull annotation

src/main/java/com/epam/aidial/deployment/manager/model/deployment/InferenceDeployment.java
```diff
@@ -17,4 +18,14 @@
...
 public class InferenceDeployment extends Deployment {
     private String modelFormat;
+
+    /**
+     * System-computed serving capability, detected from the HuggingFace model metadata at
+     * create/update time. Read-only over the API. Null until detection has run.
+     *
+     * <p>Excluded from config export/import JSON ({@link JsonIgnore}): it is derived, not authored,
+     * and is re-detected when an imported deployment is created.
+     */
```
Pasichniuk, line 28: Info belongs to inference deployments spec

src/main/java/com/epam/aidial/deployment/manager/model/deployment/InferenceDeployment.java
```diff
@@ -17,4 +18,14 @@
...
+    @JsonIgnore
```
Pasichniuk, line 29: Project uses MixIns to control what to blacklist from export files. See DeploymentExportMixIn.java

specs/024-model-serving-capability/checklists/requirements.md
```diff
@@ -0,0 +1,39 @@
+# Specification Quality Checklist: Model Serving Capability API
```
Pasichniuk, line 1: All spec-kit files except spec and research are not really useful after the feature is implemented. Recently we weren't commiting them.

CLAUDE.md
dearsimanovich-epam: Good catch — addressed in #379: removed the auto-generated `## Active Technologies` / `## Recent Changes` sections from CLAUDE.md entirely (the speckit agent-context script added them; the "Added …" claims were indeed false since those deps pre-existed).

src/main/java/com/epam/aidial/deployment/manager/web/dto/deployment/InferenceDeploymentDto.java
dearsimanovich-epam: Addressed in #379: trimmed the javadoc and added `@NotNull` — the value is always present now that the column is `NOT NULL DEFAULT 'NONE'` and the mapper coalesces null → `NONE`.

src/main/java/com/epam/aidial/deployment/manager/model/deployment/InferenceDeployment.java
dearsimanovich-epam: Addressed in #379: removed the descriptive javadoc; that detail lives in `specs/inference-deployments/spec.md`.

src/main/java/com/epam/aidial/deployment/manager/model/deployment/InferenceDeployment.java
dearsimanovich-epam: Thanks — didn't know about the MixIn convention. Addressed in #379: dropped `@JsonIgnore` and added `getInferenceTask()` to `DeploymentExportMixIn` instead, with a new `DeploymentExportMixInTest` case proving the field is stripped from the export payload.

specs/024-model-serving-capability/checklists/requirements.md
dearsimanovich-epam: Addressed in #379: kept only `spec.md` and `research.md` under `specs/024-model-serving-capability/`; removed plan.md, tasks.md, data-model.md, quickstart.md, contracts/, and checklists/.

PR #378, opened 2026-06-30
feat: auto-provision docker pull secrets for trusted-registry images

oleksii-donets:
### Applicable issues

- fixes #380

### Description of changes

At deploy time, when a deployment's in-scope image comes from a configured credentialed registry (the primary registry or a `TRUSTED_PRIVATE_REGISTRIES` entry with `BASIC` auth), the system now auto-provisions a `kubernetes.io/dockerconfigjson` pull secret and injects an `imagePullSecrets` reference into the generated CRD — so administrators configure trusted registries once instead of hand-creating pull secrets and patching service accounts.

- **In-scope images:** the deployment image for Knative image-based deployments (MCP/Interceptor/Adapter/Application); the chained text-classification transformer image for inference deployments. Predictor and NIM are out of scope.
- **Provisioning** mirrors the existing sensitive-env-var secret pattern: a per-deployment secret, created via `K8sClient`, registered as a disposable `SECRET`, and cleaned up on undeploy.
- **Least-privilege:** the deploy-time secret carries credentials only for the registry/registries serving the deployment's own images. The build-pipeline `RegistryService.dockerConfig()` aggregate (all registries) is intentionally left unchanged, since a build can pull base images from several registries.
- **Additive & gated:** public/unconfigured images are untouched; controlled by `AUTO_PULL_SECRET_ENABLED` (default `true`).

Capability specs (kubernetes-manifests, deployments, inference-deployments) and `docs/configuration.md` updated. Verified with `./gradlew testFast` and `checkstyleMain`/`checkstyleTest`.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

CLAUDE.md
```diff
@@ -72,8 +72,11 @@ Where `<slug>` is the directory name under `specs/` that the feature belongs to
...
 
 ## Active Technologies
+- Java 21 + Spring Boot 4.0.6 (Spring Framework 7); Fabric8 Kubernetes Client 7.5.2 + Knative Client 7.5.2; KServe CRD model (`io.kserve.serving.v1beta1`); Jib `ImageReference` for image parsing; MapStruct 1.6.3; Lombok (025-auto-pull-secrets)
+- No new schema. Reuses the existing `DisposableResource` table (`K8sResourceKind.SECRET`). Multi-vendor (H2/PostgreSQL/SQL Server) unaffected — no Flyway migration, no `generateDbSchema` run required. (025-auto-pull-secrets)
 - Java 21 + Spring Boot 4.0.6 (Framework 7), Hibernate ORM 7 + Envers, MapStruct 1.6.3, Flyway, Fabric8/KServe client, Lombok (024-model-serving-capability)
 - Relational, multi-vendor (H2 / PostgreSQL / SQL Server) via Flyway migrations; `inference_deployment` table (024-model-serving-capability)
 
 ## Recent Changes
+- 025-auto-pull-secrets: Added Java 21 + Spring Boot 4.0.6 (Spring Framework 7); Fabric8 Kubernetes Client 7.5.2 + Knative Client 7.5.2; KServe CRD model (`io.kserve.serving.v1beta1`); Jib `ImageReference` for image parsing; MapStruct 1.6.3; Lombok
```
Pasichniuk: Redundant change. We definitely need to teach spec-kit to avoid doing such changes or committing them

src/main/java/com/epam/aidial/deployment/manager/service/RegistryPullSecretProvisioner.java
```diff
@@ -0,0 +1,98 @@
...
+        // Narrowed to the registries that serve THIS deployment's images — not the full build-time
+        // dockerConfig() aggregate. Empty when no in-scope image matches a credentialed registry.
+        var dockerConfig = registryService.dockerConfigForImages(inScopeImages);
+        if (dockerConfig.isEmpty()) {
+            log.debug("No in-scope image of deployment '{}' matches a credentialed configured registry; "
+                    + "no pull secret provisioned", deploymentId);
+            return Optional.empty();
+        }
+
+        var secretName = K8sNamingUtils.generateUniqueName(deploymentId, PULL_SECRET_NAME_SUFFIX);
```
Pasichniuk, line 90: **Superseded pull secrets are never cleaned up (resource leak).**

`provisionForDeployment` runs inside `prepareServiceSpec`, which is invoked by both `deploy()` and `rollingUpdate()` (`AbstractDeploymentManager` lines 131 and 220). Every call mints a **new, randomly-named** secret via `generateUniqueName(deploymentId, "pull")` and registers it `STABLE` — but nothing ever marks the *previous* pull secret `TO_CLEANUP`:

- `undeploy()` only marks the service + CNP (`markKnativeServiceResourceForCleanup`), not `SECRET` resources.
- There is no pull-secret analog of `cleanupSecrets()`.

So for any deployment served from a credentialed registry, **each config update / `updateImageDefinitionForDeployments` / redeploy leaves behind an orphaned `STABLE` dockerconfigjson secret** (a K8s object + a `disposable_resource` row), only reclaimed on full deployment deletion.

This breaks the symmetry with the env-secret pattern this mirrors: env secrets use a **stable** name persisted on the deployment and are actively superseded via `cleanupSecrets()` on change. The running workload is unaffected (the new spec references the new secret), so this is a hygiene/leak issue rather than a correctness break.

Suggested fix: either mark the prior pull secret `TO_CLEANUP` before creating the new one, or use a deterministic per-deployment name + create-or-replace (verify that doesn't break Knative revision rollback, which may still reference the old secret name). Note the rollback path is already self-healing — the `saveK8sResources` row commits `TEMPORARY` via `REQUIRES_NEW`, so a rolled-back deploy leaves a cleanable secret; only *successful* updates leak.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/KnativeDeploymentManager.java
```diff
@@ -130,6 +139,17 @@ protected Service prepareServiceSpec(Deployment deployment) {
...
+        applyImagePullSecret(service, pullSecretName);
+        return service;
+    }
+
+    private void applyImagePullSecret(Service service, String pullSecretName) {
+        if (StringUtils.isBlank(pullSecretName)) {
+            return;
+        }
+        service.getSpec().getTemplate().getSpec().setImagePullSecrets(
+                List.of(new LocalObjectReferenceBuilder().withName(pullSecretName).build()));
```
Pasichniuk, line 155: **Missing null-guards (inconsistent with the Inference variant).**

This chains four getters unguarded, whereas `InferenceDeploymentManager.applyTransformerImagePullSecret` defensively checks `getSpec()`/`getTransformer()` for null. In practice `knativeManifestGenerator.serviceConfig` always builds template+spec so this won't NPE today, but the asymmetry is worth a guard (or a short comment noting the invariant).

Pasichniuk (commented):
Two findings from the review (leak + defensive-coding). See inline comments.

CLAUDE.md
oleksii-donets: Reverted in 08c61450 — dropped the three auto-appended `025-auto-pull-secrets` lines from `## Active Technologies` / `## Recent Changes`. Note the append comes from speckit's vendored `update-agent-context` script, which CLAUDE.md marks off-limits (gets overwritten on `specify update`), so the practical mitigation is stripping these during PR prep rather than editing the vendored script.

src/main/java/com/epam/aidial/deployment/manager/service/RegistryPullSecretProvisioner.java
oleksii-donets: Fixed in 08c61450. Went with your suggestion #1 (mark prior `TO_CLEANUP` before creating the new one) rather than a deterministic name + create-or-replace, to avoid the Knative revision-rollback caveat you flagged. Since the pull-secret name isn't persisted on the deployment, `provisionForDeployment` now self-supersedes: before registering the new secret it queries `getAllByGroupId` and marks prior `SECRET` resources whose name matches the `-pull-<random>` tail as `TO_CLEANUP` (env secrets use `-envs-`, so they're left untouched). Runs before the new secret is registered — its name matches the same pattern. Same-transaction `TO_CLEANUP` + async delete is exactly the env-secret behavior. Added a provisioner test asserting the prior pull secret is superseded while a co-located env secret is not, and a supersession scenario in the kubernetes-manifests spec.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/KnativeDeploymentManager.java
oleksii-donets: Fixed in 08c61450 — `applyImagePullSecret` now guards `getSpec()`/`getTemplate()`/`getSpec()` for null, matching `InferenceDeploymentManager.applyTransformerImagePullSecret`. As you noted it can't NPE today (the Knative manifest generator always builds template+spec), but this restores parity.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #381, opened 2026-07-01
fix: redeploy running deployments on any spec field change (#364)

### LINKED ISSUE #364: [Deployments] All Running deployments are still using the old configuration after changing Command or/and Arguments and after Save & redeploy
### Name and Version

Admin: [FE]0.18.0-dev.106 [BE]0.18.0-dev.14

1. Run containers or model servings
2. Add Arguments or/and Command
3. Click Save & redeploy

All Running deployments are using the new configuration after changing Command or/and Arguments and after Save & redeploy

### What do you see instead?

All Running deployments are still using the old configuration after changing Command or/and Arguments and after Save & redeploy

Image

### Additional information

_No response_

oleksii-donets:
## Summary

Fixes #364: changing a deployment's **Command** or **Arguments** and clicking "Save & redeploy" left the running workload on the old configuration.

Root cause: `DeploymentService.isApplicableForRollingUpdate(...)` decided whether an update to a `RUNNING` deployment needs a rolling update by **enumerating** the fields to compare, and checked `command`/`args` **only** inside the `InferenceDeployment` branch of a `switch`. Those fields live on the base `Deployment` class and are inherited by every type (mcp, adapter, application, interceptor, nim), so a command/args change on any other type was silently ignored — no rolling update, stale config.

The enumeration was fragile and already broken in more places than the reported one: `probeProperties`, `transport`, `mcpEndpointPath`, `modelFormat`, and `storageSize` also affect the generated manifest but were never compared.

## Approach

Invert the mechanism so it can't silently break again: treat **every** deployment property as redeploy-triggering **by default**, and exclude only an explicit deny-list of non-spec fields.

- Replaced the hand-maintained allow-list + subtype `switch` with `EqualsBuilder.reflectionEquals(existing, updated, NON_REDEPLOY_FIELDS)` (commons-lang3, already a dependency).
- `reflectionEquals` walks the full class hierarchy, so subclass fields (`transport`, `modelFormat`, `containerGrpcPort`, …) and any future field are covered automatically with no code change.
- Deny-list (fields that do NOT trigger a rolling update): `id`, `displayName`, `description`, `topics`, `metadata`, `status`, `url`, `createdAt`, `updatedAt`, `author`, `serviceName`, `allowedDomains`, `envs`. The last two have dedicated update paths (`allowedDomains` → Cilium network policy refresh; `envs` → the secret-aware `EnvVarChangeDetector`).

This is a strict **superset** of the previous detection — it can cause *more* rolling updates (the fix), never fewer.

## Changes

- **`service/deployment/DeploymentService.java`** — new `NON_REDEPLOY_FIELDS` constant + reflective comparison; removed now-unused `InferenceDeployment`/`NimDeployment` imports.
- **`DeploymentServiceTest.java`** — regression tests: command / args / subclass-field (`mcpEndpointPath`) changes → rolling update fires; `displayName`-only change → no rolling update (guards against over-triggering).
- **`specs/deployments/spec.md`** — documented the redeploy-by-default behavior and the excluded non-spec fields.

## Verification

- `testFast --tests DeploymentServiceTest` + `checkstyleMain`/`checkstyleTest` pass.
- Update-touching functional tests pass: `DeploymentFunctionalTest` (incl. the displayName+allowedDomains "no rolling update" assertion), `TopicFunctionalTest`, `DeploymentRollbackFunctionalTest`, `InferenceDeploymentManagerTest`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -64,6 +63,21 @@
...
+    /**
+     * Deployment fields that must NOT trigger a rolling update when changed: identity / audit
+     * metadata, system-computed fields, DIAL catalog display metadata, and fields handled by a
+     * dedicated path (env changes via {@link EnvVarChangeDetector}, {@code allowedDomains} via the
+     * Cilium network policy update). Every other field — base or subclass — is treated as
+     * redeploy-triggering by default, so new deployment properties are covered automatically.
+     *
+     * @see #isApplicableForRollingUpdate(Deployment, Deployment, boolean)
+     */
+    static final String[] NON_REDEPLOY_FIELDS = {
```
Pasichniuk, line 79: inferenceTask now triggers redeploys, but for TEXT_GENERATION/NONE it produces an identical manifest. Worth adding to NON_REDEPLOY_FIELDS list.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
```diff
@@ -64,6 +63,21 @@
 @RequiredArgsConstructor
 public class DeploymentService {
 
+    /**
+     * Deployment fields that must NOT trigger a rolling update when changed: identity / audit
```
Pasichniuk, line 67: The moment someone adds a new spec field whose type lacks a value equals() (a plain POJO without `@Data/@EqualsAndHashCode`), reflectionEquals falls back to Object identity equality — two distinct instances are never equal — so every update to any deployment would trigger a spurious rolling update. The old enumeration shared this reliance, but only over ~6 curated fields; the new surface is all fields and future ones. Nothing guards this. Worth a one-line note in the `NON_REDEPLOY_FIELDS` javadoc: "every reachable field type must implement value equals().

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
oleksii-donets: I'd keep `inferenceTask` redeploy-triggering — blanket-ignoring it reintroduces exactly the silent-stale-config bug class this PR fixes.

The key case is the `TEXT_CLASSIFICATION` boundary: a flip `NONE`/`TEXT_GENERATION` → `TEXT_CLASSIFICATION` (or back) **does** change the manifest shape — the chained transformer is added/removed, the CNP gets the chained-mode rules, and the public endpoint moves to the transformer. And this flip can happen with **no other field changing**: the field is system-computed (`enrichBeforePersist` re-detects from live HF Hub `pipeline_tag`/`architectures` on every update, before the comparison runs), so an upstream metadata edit by the model author — or a detector change between releases — is enough. With `inferenceTask` in `NON_REDEPLOY_FIELDS`, that update would persist the new task to the DB but leave the running workload on the old shape, with no rolling update.

You're right that a `NONE` ↔ `TEXT_GENERATION` flip re-applies an identical manifest. But that false positive needs upstream metadata drift *plus* an update touching nothing else redeploy-worthy, and it costs one redundant rolling update that self-heals (the stored value then matches). The false negative under the exclusion is a silent correctness bug. Given the PR's invariant ("strict superset — more rolling updates, never fewer"), the asymmetry favors keeping it.

A precise middle ground would be to special-case "redeploy only when the flip crosses the `TEXT_CLASSIFICATION` boundary", but that reintroduces the hand-maintained enumeration this PR removes — happy to do it if you feel strongly about the redundant-restart case.

src/main/java/com/epam/aidial/deployment/manager/service/deployment/DeploymentService.java
oleksii-donets: Good point — addressed in 13b84f0d. Added the note to the `NON_REDEPLOY_FIELDS` javadoc, and went one step further: since a javadoc sentence can't stop anyone, there's now a guard test (`deploymentFieldTypes_shouldAllImplementValueEquals`) that reflectively walks every type reachable from the `Deployment` hierarchy (following `@JsonSubTypes` for polymorphic fields like `Source`) and fails if any concrete project-owned type doesn't override `equals(Object)`. A future POJO without `@Data`/`@EqualsAndHashCode` breaks the build instead of silently redeploying on every update.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #390, opened 2026-07-16
fix: keep pull secret stable so scale-to-zero deployments resume

oleksii-donets:
### Applicable issues

- fixes #387

### Description of changes

Scale-to-zero deployments failed to auto-resume: the auto-provisioned docker pull secret was named with a fresh random suffix on every deploy/redeploy and the prior one condemned, so the scheduled disposable-resource cleaner could delete the secret that a still-live (immutable) Knative revision referenced. On scale-from-zero the activator's pod then couldn't pull its image and the deployment never resumed — silently: no pod logs/events and no status change (the manager watches only the Knative Service, whose `Ready` stays `True`). A manual Stop/Start recovered it only until the next idle-to-zero.

Fix — give the pull secret a **deterministic per-deployment name** and provision it via an **idempotent create-or-replace**, so the name the live/scaled-to-zero revision references always exists. The lifecycle now mirrors the `CiliumNetworkPolicy`: the plan is computed pre-commit (read-only) and the cluster write is applied in the `afterCommit` phase alongside the CRD, so a rolled-back/failed deploy never strands a mutated secret. A now-unused pull secret is condemned when an update moves the image to a public/unconfigured registry; legacy random-named secrets are migrated on the next redeploy.

- `RegistryPullSecretProvisioner`: split into read-only `plan()` + post-commit `apply()`; deterministic `generateName()`; `REQUIRES_NEW` disposable tracking
- `K8sClient`: add `createOrReplaceSecret` (Fabric8 `createOr(update)`)
- `AbstractDeploymentManager`: `PreparedService` holder + `applyPullSecretPlan` hook invoked in `deploy()`/`rollingUpdate()` afterCommit, before the CRD apply
- Knative/Inference plan pre-commit and apply post-commit; NIM unchanged (keeps its NGC secret)
- `specs/025-auto-pull-secrets`: FR-016/016a/017/018 + scale-to-zero scenario

Existing broken deployments recover on their next redeploy/Stop-Start. Verified: `checkstyle` + full `testFast` (H2) pass on JDK 21.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

src/main/java/com/epam/aidial/deployment/manager/service/RegistryPullSecretProvisioner.java
```diff
@@ -56,76 +66,154 @@ public RegistryPullSecretProvisioner(
...
 
-        // Supersede any prior pull secret for this deployment so a config-update / rolling-update does
-        // not orphan STABLE dockerconfigjson secrets (mirrors env-secret cleanupSecrets()). Must run
-        // before the new secret is registered — its name also matches the pull pattern.
-        supersedePriorPullSecrets(deploymentId, namespace);
-
-        var secretName = K8sNamingUtils.generateUniqueName(deploymentId, PULL_SECRET_NAME_SUFFIX);
+        // Deterministic per-deployment name so redeploy performs an idempotent create-or-replace of the
+        // SAME object the live workload already references — instead of minting a new random name and
+        // orphaning the old one, which a scaled-to-zero Knative revision could still reference (#387).
```
Pasichniuk: Remove GH issue reference

oleksii-donets: Done in 027a0246 — removed the issue reference here and the four other `#387` mentions in main/test code comments (kept the references in `specs/`, where cross-referencing is the convention).

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
```diff
@@ -137,6 +140,9 @@ public Deployment deploy(String id) {
                 @Override
                 public void afterCommit() {
                     try {
+                        // Provision the pull secret before the CRD so the revision's pods can pull;
+                        // post-commit, alongside the CRD apply, mirroring the CiliumNetworkPolicy.
+                        applyPullSecretPlan(deployment, pullSecretPlan);
```
Pasichniuk, line 145: Minor / documentation suggestion: worth a one-line comment here noting that the STABLE pull secret is **intentionally retained** across a failed deploy.

On the failure path below (`catch` → `markDisposableResourcesForCleanup(...)`), the service and CNP are condemned but the pull secret provisioned by `applyPullSecretPlan` is left `STABLE`. That's deliberate and correct — `createOrReplaceSecret` is idempotent, so a retry/redeploy reuses the same object in place (no orphan), and it's reclaimed with the rest of the group on full delete — but the asymmetry with the CNP cleanup is easy to misread as a leak. A short comment would preempt the question.

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (dismissed):

src/main/java/com/epam/aidial/deployment/manager/service/deployment/AbstractDeploymentManager.java
oleksii-donets: Done in 31e78200 — added a comment on the failure path noting the pull secret is intentionally left STABLE (verified `markDisposableResourcesForCleanup` only condemns the service resource and the CNP; SECRET-kind rows are untouched).

ai-dial-actions:
<h1>Dependency Review</h1>
None

Pasichniuk (approved):

PR #395, opened 2026-07-24
feat: scrape gpu metrics

Pasichniuk:
### Applicable issues

- fixes #388 

### Description of changes

- **Wire real GPU telemetry into the metrics snapshot via the NVIDIA DCGM exporter.** New `GpuMetricsReader` discovers dcgm-exporter pods co-located on the deployment's nodes and scrapes them through the API-server pod proxy; `GpuMetricsCollector` parses `DCGM_FI_DEV_{GPU_UTIL,FB_USED,FB_FREE}`, filters by `namespace`/`pod` labels, and aggregates per pod (Σ memory, avg util). Collection is gated on `nvidia.com/gpu > 0` (`GpuRequest`), populating per-pod `gpuUtilization`/`gpuMemoryBytes` + the new `gpuMemoryTotalBytes` and flipping `resources.gpu` to available.
- **Graceful degradation preserved.** When the exporter is absent/unreachable, GPU collection is disabled, or the deployment requests no GPU, `resources.gpu` is reported unavailable with a distinct reason and every other block is unaffected — the snapshot never fails.
- **Operator configuration.** Added `app.metrics.scrape.gpu.*` properties (`MetricsScrapeProperties`, `application.yml`) with matching `docs/configuration.md` entries; `PodInfo` now carries `nodeName` for the co-location join and `ResourceMetricsDto` exposes the GPU fields.
- **Test coverage + fixtures.** New `GpuMetricsCollectorTest`, extended `DeploymentMetricsServiceTest` and `DeploymentMetricsFunctionalTest`, and a synthetic `dcgm.txt` exposition fixture (multi-GPU pod, cross-namespace noise).
- **Spec/docs hygiene.** `deployment-metrics` and `026-gpu-metrics` specs updated to mark GPU telemetry implemented and drop stale PoC framing; filed and cross-linked follow-ups [#396](https://github.com/epam/ai-dial-admin-deployment-manager-backend/issues/396) (time-range API) and [#397](https://github.com/epam/ai-dial-admin-deployment-manager-backend/issues/397) (labelSelector-scoped query).

### Checklist


src/main/java/com/epam/aidial/deployment/manager/kubernetes/K8sClient.java
```diff
@@ -97,11 +100,43 @@ public PodResource getPodResource(String namespace, String podName) {
...
+        if (podNames == null || podNames.isEmpty()) {
+            return List.of();
+        }
+        var pending = podNames.stream()
+                .map(podName -> Map.entry(podName, CompletableFuture.supplyAsync(
+                        () -> client.raw(buildProxyUri(namespace, podName, port, metricsPath)), scrapeExecutor)))
+                .toList();
+        var bodies = new ArrayList<String>(pending.size());
+        for (var entry : pending) {
+            awaitScrape(entry.getValue(), namespace, entry.getKey(), port, timeoutMs).ifPresent(bodies::add);
```
oleksii-donets: The futures are submitted concurrently, but each is reaped with a **fresh, full `timeoutMs`** — so if several exporters hang, total wall time is N×timeout (e.g. 3 nodes × 10 s default = 30 s inside a synchronous, cacheable request), contradicting the javadoc's "total latency is ~one timeout rather than the sum".

Suggest a single shared deadline:

```java
var deadlineNanos = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(timeoutMs);
var bodies = new ArrayList<String>(pending.size());
for (var entry : pending) {
    var remainingMs = TimeUnit.NANOSECONDS.toMillis(deadlineNanos - System.nanoTime());
    awaitScrape(entry.getValue(), namespace, entry.getKey(), port, Math.max(remainingMs, 0)).ifPresent(bodies::add);
}
```

(`get(0, …)` throws `TimeoutException` immediately for a not-yet-done future, which is exactly the desired expired-deadline behaviour.) Low impact in practice — GPU deployments usually span 1–2 nodes — but the doc and behaviour should agree.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/metrics/GpuMetricsReader.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        try {
+            var exporterPodNames = k8sClient.getPods(namespace, labels).getItems().stream()
+                    .filter(pod -> pod.getSpec() != null && nodeNames.contains(pod.getSpec().getNodeName()))
+                    .map(Pod::getMetadata)
+                    .filter(Objects::nonNull)
+                    .map(ObjectMeta::getName)
+                    .filter(StringUtils::isNotBlank)
+                    .toList();
+            return k8sClient.scrapePodMetrics(namespace, exporterPodNames, gpu.getPort(),
```
oleksii-donets, line 74: Exporter pods are filtered only by `spec.nodeName` — there is no pod-phase filter and no per-node dedup. If **two exporter pods coexist on one node** (DaemonSet rolling update overlap, or an over-matching selector), both get scraped and `GpuMetricsCollector` sums the same GPU series twice — doubling `gpuMemoryBytes`/`gpuMemoryTotalBytes` (utilization stays correct only because it's averaged).

Suggest keeping at most **one Running exporter per node** (newest wins — during a rollout the newer pod is the DaemonSet's current replacement):

```java
var exporterPodNames = k8sClient.getPods(namespace, labels).getItems().stream()
        .filter(pod -> pod.getSpec() != null && nodeNames.contains(pod.getSpec().getNodeName()))
        .filter(pod -> pod.getStatus() != null && "Running".equals(pod.getStatus().getPhase()))
        .filter(pod -> pod.getMetadata() != null && StringUtils.isNotBlank(pod.getMetadata().getName()))
        .collect(Collectors.toMap(
                pod -> pod.getSpec().getNodeName(),
                Function.identity(),
                (left, right) -> newestOf(left, right)))
        .values().stream()
        .map(pod -> pod.getMetadata().getName())
        .toList();
```

The Running filter also stops Pending/Failed exporter pods from burning scrape-timeout budget. (An alternative fix is deduping series by the `UUID`/`gpu` label in the collector — either way, one of the two layers should guard against double-counting.)

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
```diff
@@ -131,31 +140,90 @@ private ServingMetricsResult collectServingMetrics(Deployment deployment, Deploy
...
+        }
+        if (allPods.isEmpty()) {
+            availability.put(AVAILABILITY_RESOURCES_GPU, AvailabilityStatus.unavailable(REASON_GPU_NO_PODS));
+            return podUsages;
+        }
+        var podNames = allPods.stream().map(PodInfo::getName).collect(Collectors.toSet());
+        var nodeNames = allPods.stream().map(PodInfo::getNodeName).filter(StringUtils::isNotBlank).collect(Collectors.toSet());
+        var gpuByPod = gpuMetricsCollector.collect(manager.getNamespace(), podNames, nodeNames);
+        if (gpuByPod.isEmpty()) {
+            availability.put(AVAILABILITY_RESOURCES_GPU, AvailabilityStatus.unavailable(REASON_GPU_UNAVAILABLE));
```
oleksii-donets, line 199: If pods exist but none has a `nodeName` yet (all Pending), `nodeNames` is empty, the collector returns an empty map without any scrape, and the reason reported is *"GPU telemetry unavailable (DCGM exporter not present or unreachable?)"* — misdiagnosing the cause. The real cause is "pods not scheduled yet"; a distinct reason would keep the operator-facing diagnostics as precise as the rest of the FR-006 reason matrix:

```java
if (nodeNames.isEmpty()) {
    availability.put(AVAILABILITY_RESOURCES_GPU,
            AvailabilityStatus.unavailable("GPU usage not available until pods are scheduled onto nodes"));
    return podUsages;
}
```

This also skips the pointless exporter list call for that case.

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentMetricsFunctionalTest.java
```diff
@@ -255,10 +318,38 @@ private void stubPods(String labelKey, String serviceName, Pod... pods) {
...
+        var servicePodList = new PodList();
+        servicePodList.setItems(List.of(servicePod));
+        when(svcOp.inNamespace(NAMESPACE)).thenReturn(svcOp);
+        when(svcOp.withLabels(Map.of(KSERVE_SERVICE_LABEL, serviceName))).thenReturn(svcOp);
+        when(svcOp.list()).thenReturn(servicePodList);
+
+        var dcgmPodList = new PodList();
+        dcgmPodList.setItems(List.of(exporterPods));
+        when(svcOp.inNamespace(dcgmNamespace)).thenReturn(dcgmOp);
+        when(dcgmOp.withLabels(Map.of("app", "nvidia-dcgm-exporter"))).thenReturn(dcgmOp);
```
oleksii-donets: Nit: this hardcodes the default selector, silently coupling the stub to `application.yml` — if the default `app.metrics.scrape.gpu.pod-label-selector` ever drifts, the stub stops matching and the test degrades to the "no exporter" path rather than failing loudly. Consider autowiring `MetricsScrapeProperties` and deriving the label map from `getGpu().getPodLabelSelector()` (and the namespace from `getGpu().getNamespace()`), so the stubs always follow the configuration the reader actually uses.

src/main/java/com/epam/aidial/deployment/manager/kubernetes/K8sClient.java
Pasichniuk: Done

src/main/java/com/epam/aidial/deployment/manager/kubernetes/metrics/GpuMetricsReader.java
Pasichniuk: Done

src/main/java/com/epam/aidial/deployment/manager/service/deployment/metrics/DeploymentMetricsService.java
Pasichniuk: Done

src/test/java/com/epam/aidial/deployment/manager/functional/tests/DeploymentMetricsFunctionalTest.java
Pasichniuk: Done

oleksii-donets (approved):

PR #398, opened 2026-07-24
fix: schedule chained transformer on the same node pool as predictor

oleksii-donets:
### Description of changes

For KServe inference deployments with a chained transformer (text-classification task), pool scheduling primitives (nodeSelector / affinity / tolerations) resolved from the deployment's `nodePoolId` were applied only to the predictor block — the transformer pod could be scheduled on any node pool.

- `InferenceManifestGenerator`: added `applyTransformerPoolPrimitives` (called in the chained branch after the transformer block is built) that projects the same primitives onto the transformer: nodeSelector replaced (defensive copy), affinity replaced, pool tolerations appended — mirroring the predictor projection. Renamed the existing method to `applyPredictorPoolPrimitives`. No pool selected → transformer untouched.
- Tests: chained manifest asserts the transformer carries the pool's nodeSelector/affinity/tolerations alongside the predictor; empty primitives don't fabricate transformer scheduling fields; non-chained manifests get no transformer block.
- Specs updated: kubernetes-manifests (new requirement bullet + scenario), 019-explicit-pool-scheduling FR-006, 021-inference-task-transformer assumptions.

### Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/main/java/com/epam/aidial/deployment/manager/service/manifest/InferenceManifestGenerator.java
```diff
@@ -247,6 +252,36 @@ private void applyPoolPrimitives(MappingChain<Predictor> predictorChain, PoolSch
...
 
+    // Mirrors applyPredictorPoolPrimitives so a chained transformer schedules onto the same node
+    // pool as the predictor. Kept as a separate typed method: the generated predictor/transformer
+    // CRD types share no common interface, and MappingChain<Predictor>/MappingChain<Transformer>
+    // have the same erasure, so an overload would not compile.
+    private void applyTransformerPoolPrimitives(MappingChain<InferenceService> config, PoolSchedulingPrimitives primitives) {
+        if (primitives == null || primitives.isEmpty()) {
+            return;
+        }
+        var transformer = config.get(InferenceMappers.SERVICE_SPEC_FIELD)
```
Pasichniuk, line 264: **Consider aligning the defensive copy with the predictor path.**

Here the transformer copies the map (`new LinkedHashMap<>(primitives.nodeSelector())`), but `applyPredictorPoolPrimitives` assigns the reference directly:

```java
predictorChain.data().setNodeSelector(primitives.nodeSelector());
```

`PoolSchedulingPrimitives.of()` wraps `pool.getNodeSelector()` straight from the `NodePoolProperties.PoolConfig` singleton, which is shared across every deployment that uses the pool. So the predictor stuffs that shared, long-lived config map directly into a per-deploy manifest. Not an active bug today (the manifest map is only read during serialization), but it's a latent aliasing hazard and now inconsistent with the transformer.

The transformer's copy here is the safer pattern — suggest applying the same `new LinkedHashMap<>(...)` on the predictor path so both are consistent and neither aliases the shared pool config. (Affinity/tolerations are fine — the converter produces fresh objects each call.)

oleksii-donets: Good catch — confirmed `PoolSchedulingPrimitives.of()` passes `pool.getNodeSelector()` through by reference from the `NodePoolProperties` singleton. Aligned the predictor path with the transformer's defensive copy in 203eef07.

Pasichniuk (approved):

